PR #93, opened 2015-07-04
market: add a twitter share button

carver:
Market names that are longer than 83 characters will be too long to tweet. For now, we just rely on the user to shorten the tweet as needed.

The format of the tweet is currently:
"[market name] \#CashPrediction [market url] via @AugurProject"

Note that when testing on localhost, twitter won't url-shorten, so the tweet will typically be too long. You can modify the domain to anything else to see the real tweet length.

### DIFF
--- a/app/components/Market.jsx
+++ b/app/components/Market.jsx
@@ -93,6 +93,11 @@ var Market = React.createClass({
         </div>
         <div className='row'>
           <div className='col-xs-12'>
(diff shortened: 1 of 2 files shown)


tinybike:
Very cool :)

tinybike:
Follow-up thought: should we make this always use a link for client.augur.net (or demo.augur.net)?  If the domain is localhost, it won't lead anywhere when users click on it.

Issue #101, opened 2015-07-07
Trap and properly message user when etherbase (coinbase) account is not set


scottzer0:
This may not always be an issue but in the name of robust design should probably be implemented at least as a console error log and halt.

carver:
I'll start in on this

PR #105, opened 2015-07-07
Show help modal if user is missing --etherbase flag

carver:
Edit: Ok, this is ready to go

Should take care of #101 

I'm not totally happy with the message itself, but it should get the concept across for sufficiently advanced users.  Any suggestions on wording?

AugurApp render could use a cleanup refactor.  That will go in a separate PR.


carver:
@scottzer0 @tinybike could you review please?

PR #108, opened 2015-07-09
Double startup speed - Render MarketPane less with shouldComponentUpdate

carver:
Keep track of relevant market changes, and only update if necessary.  I tried the immutable-js route with the PureRenderMixin, but there is market mutation everywhere, and parts of the system seem to depend on that.

In tests on my local machine before this change, I would typically see markets first render, and the whole page complete, in about 12 and 16 seconds, respectively.  After this change, I was seeing about 5 and 7.

There are still another couple seconds to shave off in React render sections, then we'll have to start looking elsewhere for the biggest wins.

### DIFF
--- a/app/actions/MarketActions.js
+++ b/app/actions/MarketActions.js
@@ -187,6 +187,13 @@ var MarketActions = {
 
   updateMarket: function(market, supplement) {
 
+    // Calculate market properties before dispatch (seems to belong in a Market class)
+    if (!market.volume && market.volume !== 0) {
+      market.volume = _.reduce(market.outcomes, function(volume, outcome) {
+        if (outcome) return volume + parseFloat(outcome.volume);
+      }, 0);
+    }
+
     this.dispatch(constants.market.UPDATE_MARKET_SUCCESS, {market: market});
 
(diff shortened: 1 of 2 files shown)


scottzer0:
nice work here!

niran:
I strongly support immutable-js if you're trying to head in that direction.

carver:
Cool, I'll try to fit it in. Maybe next.



Sent on a Sprint Samsung Galaxy S® 5


-------- Original message --------

niran:
My only concern is that it's probably an extensive change.

carver:
Agreed.  Immutable-js may cause more instability in the short term, but hopefully would have long-term benefits.  Eventually, I’ll head down the path again and see how far I can get.

The first time I tried, it started breaking in unclear ways (eg~ page didn’t load, without any console errors).  I was trying to do an incremental update and just fix everywhere that broke, but that seems to be an ineffective strategy.  Instead, I’ll try to follow through everywhere that uses the market store, and update it.

I am getting married in a couple weeks, so between that and a full-time job, I’ll probably be pretty silent until mid-August.

niran:
Congrats!

PR #111, opened 2015-07-10
shouldComponentUpdate must strictly return boolean

carver:
This was causing warnings about returning undefined.

I'm away from my dev machine, but this should fix it.


tinybike:
Yep, fixes it for me!  Merging.

PR #193, opened 2015-09-29
fixes typo res => err

davidtrogers:
Hi, this fixes a small typo: `res is undefined`


tinybike:
Actually I think it should just be `txhash` there (defined in the enclosing scope), since `err` is an error object and doesn't have a `txHash` property.  Thanks for bringing this to my attention; closing and pushing a fix.

PR #208, opened 2015-10-26
Add in Toggle Market Button to appear on markets page

Frankenmint:
Created in response to: https://github.com/AugurProject/augur/issues/189

### DIFF
--- a/app/actions/MarketActions.js
+++ b/app/actions/MarketActions.js
@@ -51,23 +51,46 @@ var MarketActions = {
     }
   },
 
-  parseCache: function (cached) {
+  
+//Toggles Mature Markets Switch from Branch.jsx Control Button
+  mMarkets: function() {
+    this.flux.actions.market.mMarkets = true;
+  },
+  
+
+  parseCache: function (cached, mMarkets) {
     var self = this;
     var block = self.flux.store('network').getState().blockNumber;
     var account = self.flux.store('config').getAccount();
     var branchId = self.flux.store('branch').getCurrentBranch().id;
     var blackmarkets = blacklist.markets[augur.network_id][branchId];
     var markets = {};
+    
+    mMarkets = self.flux.actions.market.mMarkets;
+    
+    var mktsToggler; 
+
     async.eachSeries(cached, function (thisMarket, nextMarket) {
       var marketId = abi.bignum(thisMarket._id);
+      var endDate = thisMarket.endDate;
+       
+      //Filters markets to either active or matured on mktsToggler
+      if (mMarkets) mktsToggler = endDate - block > 0 ? true : false;
+      else mktsToggler = endDate - block < 0 ? true : false;
+
+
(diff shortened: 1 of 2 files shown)


tinybike:
Thanks for taking the time to put this together.  It's a good start, but there are a few problems that jump out at me:

- The toggle switch could use some kind of label.  Right now it's hard to tell what it does prior to clicking on it.
- Clicking the toggle triggers a full market reload.  This isn't obviously a problem right now, because the total number of markets is relatively small and it's on a hosted node.  But, it would be much more efficient to just filter the markets that are already in the data store, rather than re-fetching all the markets from scratch.
- `parseCache` is not a good place to do the filtering, because (as its name suggests) this method is only invoked when loading data from the MongoDB.  To make this work for local (direct-to-geth) connections, you'd need to either create a second equivalent block of code somewhere in the loadMarkets code, or (preferably) insert this logic somewhere else where it doesn't care about where the data came from and doesn't couple the toggle to the initial loading logic.
- It looks like you're using `mMarkets` [both as a method and as a variable](https://github.com/Frankenmint/augur/commit/6baf583c7e5604edf5c9ca60064684816242b1ba#diff-4e1685aa4d95f364c4fd5642014f8c8bR57) (to store the toggle's state).  This is confusing and you should use multiple labels instead of just one.  A bigger issue is that state shouldn't be saved in the actions -- usually you should use the stores for this, but if you need to it's ok to save some state info within the component (if the state does not need to persist after the component unloads).  But don't store it in the action!  Stuff like this `window.flux.actions.market.mMarkets = true;` should **never ever ever** be happening in the components.

Leaving this PR open for now.  (This code will likely need a rewrite before it's ready to merge.)

Frankenmint:
* what about something like an open sign/closed sign icon set?
* so I know what you're saying, I have different code that was based on what you mentioned before but every time an action happens it uses the parsecache to update everything so when I used my function to toggle, it would still randomly run the parsecache function and erase what I filtered away from the store.  I'd prefer it I could scrap this portion or make it function for a local node which seems ideal long term.  

````javascript
toggleMarketsOn: function() {
  var self = this;
  var storedMarkets = self.flux.store('market').getState().markets; // from marketStore
  var currentPeriod = 191; //flux.store('branch').getState().currentBranch.currentPeriod;
  var markets = {};  

  for (var market in storedMarkets) {

      var marketExpires = storedMarkets[market].tradingPeriod.toNumber();

      if (!storedMarkets.hasOwnProperty(market)) continue;
      if (currentPeriod >= marketExpires) delete storedMarkets[market];
      markets = storedMarkets;
    
  }
 // save markets to MarketStore
    self.dispatch(constants.market.LOAD_MARKETS_SUCCESS, { markets:  markets }); 
    console.log(Object.keys(markets).length + " Markets Found");
}
   ```


* as for replacing window. I'm reading what you've shared now, thx!

PR #285, opened 2016-02-12
282 new markets page design

thinkloop:
Changed some global styles and such, would like to get this in sooner rather than later as it might help other sections


tinybike:
Looking sharp :)

PR #607, opened 2016-05-17
Add tests

JohnDanz:
Ready to be reviewed for the first pass of Unit Tests, keep in mind this will need to be updated throughout the beta to stay current.


thinkloop:
The tests look great! Would you like to give us a tour on the call today?

A minor thing, the rest of the code uses tabs, but I noticed the tests use spaces, possible to change the spaces to tabs?

stephensprinkle:
To your point re: tabs vs spaces — would we want to implement [editorconfig](http://editorconfig.org/) to define the code styles?

PR #614, opened 2016-05-24
Add tests

JohnDanz:
This should fix the default() function position in all files where it had been moved to the bottom of the page to satisfy an eslint rule instead of at the top for readability reasons. The rule has been disabled (no-use-before-defined).


thinkloop:
looks good! merging

thinkloop:
leaving merge for sprinkle

Issue #620, opened 2016-06-03
Update create market form with new fields and liquidity step


thinkloop:
The logic is located in `/src/modules/create-market`

Part of this issue will be to implement the new `generateOrderBook` function jack added as an action: https://github.com/AugurProject/augur.js/blob/master/src/index.js#L854 

Part will be to update the `create-market-form` selector

stephensprinkle:
TODO:

- [ ] Update `create-market-form` selectors
- [ ] Implement `generateOrderBook`

stephensprinkle:
PR #652 resolved this issue.

PR #624, opened 2016-06-14
#620 create market

stephensprinkle:
This PR resolves the issue of not being able to create a market.  Additionally included in this PR is the ability to create a market with a subsequent order book created via `generateOrderBook`.

NOTE -- Full test coverage is missing.  I will be building those out and will submit an additional PR for the inclusion of those.  Once tests are complete, issue #620 will thus be complete + closed.


tinybike:
Merging this to master so it can be tested in tandem with the new UI trading code.

PR #705, opened 2016-07-19
684 account funding

JohnDanz:
Adds funding to registration again, Must be reviewed as I'm having trouble adding the funding to the transactions in the UI.


JohnDanz:
Fixed transactions. Found issue with funding. If you register your account but don't login then the funding will fail. it appears to be looking for the local geth node account since the error message is "account is locked." This may be an issue in the backend that needs to be changed to use the augur.web.account object instead.

PR #714, opened 2016-07-22
Import straight from AURC/src (quick fix)

priecint:
#713


thinkloop:
I think we should fix it properly with https://github.com/AugurProject/augur-ui-react-components/issues/62

tinybike:
Agree that https://github.com/AugurProject/augur-ui-react-components/issues/62 is a cleaner fix; closing this PR.

@priecint, is it correct that fixing https://github.com/AugurProject/augur-ui-react-components/issues/62 will also fix npm link?  If so, I think we should bump it way up the priority list -- npm link is awfully helpful when it's working.

PR #717, opened 2016-07-25
`loadMarketsInfo` expected an array as the first argument

stephensprinkle:


thinkloop:
Hey sprinkle,

Loading the market here is not need since we have a block listener for new markets: https://github.com/AugurProject/augur/blob/master/src/modules/app/actions/listen-to-updates.js#L32

Better to get rid of it

stephensprinkle:
Ah, perfect.  Thanks for pointing that out!

I'll remove that call + submit a new PR.

PR #719, opened 2016-07-25
Account page

JohnDanz:
Adds the ability to download the account and also shows/hides pieces of the account page depending on which type of account is logged in, Hosted vs Local Geth Node.


thinkloop:
Hey @JohnDanz, 

The link url should just be a string passed to aurc, then aurc renders a <Link /> component. See some other link examples. We should never create html elements in augur.

JohnDanz:
We don't actually render the link, it simply creates one, sets the values that need to be on it to force a download, then clicks it to activate it.

thinkloop:
much better!

PR #740, opened 2016-08-09
204 my markets

stephensprinkle:


tinybike:
Hmm, I'm getting the following error during the initial load:
```
load-login-account.js:56: Uncaught TypeError: (0 , _updateAccountTradesData.updateAccountTradesData) is not a function
```
Any idea what might be causing this?

tinybike:
Addendum: that error only appears when using a local unlocked account.  When I use a client-side account (i.e., normal register/login) it works ok.  Maybe there's something in updateAccountTradesData looking for a `loginAccount` field that isn't populated if you use a local account?  (Trading in general is still broken on local accounts AFAIK, possibly for the same reason.)

tinybike:
In case you don't have geth setup to test this: local unlocked accounts have `loginAccount.id` set (to the address) but `loginAccount.address` and `loginAccount.keystore` are not set.  @JohnDanz is there a reason `loginAccount.address` isn't set for unlocked accounts / could that be breaking this + trading?

JohnDanz:
So it's not set because I was under the impression that we could just use id moving forward and that would work, the only reason address is still there is that i haven't gone through to make sure loginAccount.address is never referenced. So to fix the local account issue we would want to change anything looking for loginAccount.address to just use loginAccount.id instead.

PR #741, opened 2016-08-10
Transfer funds

JohnDanz:
Adds Transfer Funds to the account page.


priecint:
- display more info for transaction, e.g. amount to send and recepient
image
 - missing tests for process-transfer-funds.js and and-transfer-funds.js
 - call directly augur.js function instead of adding wrapper in augurjs.js file

PR #742, opened 2016-08-10
market detail

priecint:


tinybike:
I'm going to go ahead and merge this, however the placement of the extra info fields needs some work IMO.  (All the info fields probably shouldn't just be crammed into the top of the trade panel in a big horizontal list.)

PR #756, opened 2016-09-01
171 link row to market

stephensprinkle:


tinybike:
I'm getting the following error:
```
transactions.js:43 Uncaught TypeError: Cannot read property 'id' of undefined
```
Just a missing existence check, maybe?

tinybike:
Yep!  Was choking on the fund new account transaction (which doesn't have a `data` property).  This still needs from code in AURC to actually make the rows clickable, I assume?

PR #757, opened 2016-09-02
reset current outcome trades whenever a field is cleared

stephensprinkle:


tinybike:
Looks good!  Merging.

PR #766, opened 2016-09-08
Fixes Sign Out Link, changed when fund new account occurs

JohnDanz:
Sign Out link now brings uses back to the login page after signing out. 

the fund new account action won't be triggered until after the user hits "sign in" after registering a new account.


tinybike:
Looks good!

PR #776, opened 2016-09-20
86 select currency

JohnDanz:
Should allow the transfer of realEth and REP using the account page. This PR should be paired with the AURC branch by the same name (86-selectCurrency).


JohnDanz:
https://app.clubhouse.io/augur/story/86/add-currency-selector-for-send-cash

PR #780, opened 2016-09-23
Refresher button for the trading panel

priecint:
Added: timer, meta.ignore to actions, market data timestamps, formatTImePassed method, selectors, reducers, actions


tinybike:
@tomasprie Can you change this so that there is a visual indication that the market is refreshing, but the button isn't disabled?  (Greying out the button is fine, so long as clicking it still works.)  15 seconds is just an average; sometimes blocks arrive more rapidly than that.  Users should be able to refresh as rapidly as they want.

tinybike:
I'm going to go ahead and merge this, since it's better than no refresh button at all, but please implement the above change.

PR #797, opened 2016-10-26
Calculate how many shares a user can buy/sell

priecint:


tinybike:
@priecint Your calculation is not right.  Here is a binary market where it says 0 shares max (which is clearly wrong):

screenshot - 10262016 - 09 33 49 pm

And, here is a scalar market where it appears to just be pulling the amount in the best offer (I should be able to buy more than 3.17 given the eth I have available):

screenshot - 10262016 - 09 21 11 pm

PR #798, opened 2016-10-26
Prevent placing trade if user doesn't have enough funds

priecint:


tinybike:
@priecint This doesn't work correctly for sell orders:

screenshot - 10262016 - 10 12 11 pm

Broken for short sell too:

screenshot - 10262016 - 10 05 09 pm

priecint:
I'm using `trade.totalCost` (https://github.com/AugurProject/augur/blob/91c808375291cd7c10bb0e8eea42b040edb6b5ea/src/modules/trade/helpers/has-user-enough-funds.js#L20-L21) to calculate the total cost of all the trades and then comparing it with user funds, what am I doing wrong?

tinybike:
Only use total cost for buy/bid trades.  For sells, you should use the trading fee -- total cost is how much the user gets from the trade (not great variable naming I know...).

PR #799, opened 2016-10-27
832 update trading ui tests

JohnDanz:
Adds Trade Tests for the UI.

https://app.clubhouse.io/augur/story/832/unit-regression-tests-for-trading-ui


tinybike:
Quick comment: the id field for orders in the order book is a hex string, and should exactly match the key for the order.  (Mocks in the new tests have keys like `'order1'` and id fields like `1`.  Not a big deal, but mock structure should match the real one when possible IMO!)

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+				marketDescription: 'test scalar market',
+				outcomeName: ''
+			},
+			feePercent: {
+				value: '0.95763203714451532'
+			}
+		}
+	};
+	testState.orderBooks = {
+		'0x000000000000000000000000000000000binary1': {
```
tinybike: Suggest using valid hex strings here -- regular human-readable strings are ok too if needed, but mashups like `0xbinary1` are nonsense and probably should be avoided.

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+	const mockAugur = { augur: { ...augur }, abi: { ...abi }, constants: { ...constants } };
+	const mockLoadBidAsks = { loadBidsAsks: () => {} };
+	sinon.stub(mockLoadBidAsks, 'loadBidsAsks', (marketID, cb) => {
+		assert.isString(marketID, `didn't pass a marketID as a string to loadBidsAsks`);
+		cb();
+		return { type: 'LOAD_BIDS_ASKS' };
+	});
+
+	sinon.stub(mockAugur.augur, 'getParticipantSharesPurchased', (marketID, userID, outcomeID, cb) => {
+		cb(10);
```
tinybike: `getParticipantSharesPurchased` returns a stringified number, not a JS number (`'10'` instead of `10`).

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+		onTradeHash({ data: 'data' });
+		onCommitSent({ data: 'data' });
+		onCommitSuccess({ gasFees: '0.01450404', hash: 'testhash', timestamp: 1500000000 });
+		onCommitFailed({ code: 'error', message: 'error message' });
+		onNextBlock({ data: 'data' });
+		onTradeSent({ data: 'data' });
+		if (mockAugur.augur.trade.callCount === 1) {
+			onTradeSuccess({ sharesBought: '10', cashFromTrade: '0', unmatchedShares: '0', unmatchedCash: '0', tradingFees: '0.01', gasFees: '0.01450404', hash: 'testhash', timestamp:1500000000 });
+		} else {
+			onTradeSuccess({ sharesBought: '0', cashFromTrade: '10.00', unmatchedShares: '0', unmatchedCash: '0', tradingFees: '0.01', gasFees: '0.01450404', hash: 'testhash', timestamp:1500000000 });
```
tinybike: Number strings are inconsistent here -- `cashFromTrade` has 2 decimal points (`'10.00'`) while `sharesBought` has none (`'10`').  It's probably best to just use the same format as the real `trade` function, which does not display trailing zeros.

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+
+	sinon.stub(mockAugur.augur, 'getCashBalance', (takerAddress, cb) => {
+		cb('10000.0');
+	});
+
+	sinon.stub(mockAugur.augur, 'trade', (args) => {
+		const { max_value, max_amount, trade_ids, sender, onTradeHash, onCommitSent, onCommitSuccess, onCommitFailed, onNextBlock, onTradeSent, onTradeSuccess, onTradeFailed } = args;
+		onTradeHash({ data: 'data' });
+		onCommitSent({ data: 'data' });
+		onCommitSuccess({ gasFees: '0.01450404', hash: 'testhash', timestamp: 1500000000 });
```
tinybike: Why pass these three values to the callback, and not the others it is usually passed?

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+	});
+
+	sinon.stub(mockAugur.augur, 'getCashBalance', (takerAddress, cb) => {
+		cb('10000.0');
+	});
+
+	sinon.stub(mockAugur.augur, 'trade', (args) => {
+		const { max_value, max_amount, trade_ids, sender, onTradeHash, onCommitSent, onCommitSuccess, onCommitFailed, onNextBlock, onTradeSent, onTradeSuccess, onTradeFailed } = args;
+		onTradeHash({ data: 'data' });
+		onCommitSent({ data: 'data' });
```
tinybike: Like every other `onSent` callback, `onCommitSent` expects to receive an object `{ txHash, callReturn }`.  Why is it being passed `{ data: 'data' }` here?

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+		cb(10);
+	});
+
+	sinon.stub(mockAugur.augur, 'getCashBalance', (takerAddress, cb) => {
+		cb('10000.0');
+	});
+
+	sinon.stub(mockAugur.augur, 'trade', (args) => {
+		const { max_value, max_amount, trade_ids, sender, onTradeHash, onCommitSent, onCommitSuccess, onCommitFailed, onNextBlock, onTradeSent, onTradeSuccess, onTradeFailed } = args;
+		onTradeHash({ data: 'data' });
```
tinybike: `onTradeHash` expects a string argument (the trade hash), not an object.

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+		onCommitSuccess({ gasFees: '0.01450404', hash: 'testhash', timestamp: 1500000000 });
+		onCommitFailed({ code: 'error', message: 'error message' });
+		onNextBlock({ data: 'data' });
+		onTradeSent({ data: 'data' });
+		if (mockAugur.augur.trade.callCount === 1) {
+			onTradeSuccess({ sharesBought: '10', cashFromTrade: '0', unmatchedShares: '0', unmatchedCash: '0', tradingFees: '0.01', gasFees: '0.01450404', hash: 'testhash', timestamp:1500000000 });
+		} else {
+			onTradeSuccess({ sharesBought: '0', cashFromTrade: '10.00', unmatchedShares: '0', unmatchedCash: '0', tradingFees: '0.01', gasFees: '0.01450404', hash: 'testhash', timestamp:1500000000 });
+		}
+		onTradeFailed({ code: 'error', message: 'error message' });
```
tinybike: Middleware error objects have fields `{ error, message }`, not `{ code, message }`.

test/trade/actions/helpers/trade-test.js
```diff
@@ -0,0 +1,248 @@
...
+		'../../../../services/augurjs': mockAugur
+	});
+
+	beforeEach(() => {
+		store.clearActions();
+		mockCB.reset();
+		mockCBStatus.reset();
+	});
+
+	it('should help with sell trades', () => {
```
tinybike: The `trade` function is pretty central to trading.  So IMO we should write more than a single buy test case and a single sell test case.  Especially important is to include the relevant edge cases you can think of.  Here are a few I can think of, but this is by no means an exhaustive list:
    - What if there are no matching trade IDs (was filled at the last second before `trade` was called)?
    - What if there are a *lot* of trade IDs, maybe 50 or 100?
    - Right now trade IDs are chunked in batches of 3, so probably worth at minimum testing a number of trade IDs that is a multiple of 3 and a number that isn't.
    - Does `trade` cope gracefully with stringified number inputs?  BigNumber inputs?  JS number inputs?  (It should handle all 3; there should be test cases for each.)
    - How about if some/all of the inputs are negative?
    - How about if some/all of the inputs are null or undefined?  (If it throws an error, test to make sure it's the exact error that is expected for that test case.)
    - What happens if the `onFailed` handlers are invoked?  Does `trade` deal gracefully with these?  How is the transaction object updated if `onFailed` is called?
    - There are 4 different conditions in [async.until](https://github.com/caolan/async/blob/v1.5.2/README.md#until) that can cause the loop to terminate.  There should be at least 1 test case that covers each of these conditions.
    - What happens if `augur.getParticipantSharesPurchased` or `augur.getCashBalance` return a value that is different than expected (perhaps the user had other actions running in parallel with the trade)?  Does `trade` respond appropriately, or does it still try and use the (now outdated) values?

tinybike:
I'm going to hold off reviewing the rest of the code as I don't think this is anywhere near ready to merge.  @JohnDanz: please do a second thorough pass of these functions.

test/trade/actions/helpers/trade-test.js
JohnDanz: the functions I was checking on those callbacks do nothing with the data back and simply console log that the trade was sent. I will correct this and return a proper object.

test/trade/actions/helpers/trade-test.js
JohnDanz: I'm going to change these to human readable skins, something like `testBinaryMarketID` so that it's very clear what type of market we are looking to test against. I would be happy to change to hex strings if it is prefered but I like the human readables in this case just to make it clearer when looking at the tests.

JohnDanz:
These test are **not** complete however they should be safe to merge in for now. I will be opening new PR's for more thorough tests. Smaller PRs should be easier for the team to review and confirm that the tests are accurate and complete.

tinybike:
@JohnDanz these tests are not passing for me -- I get the following error:
```
  1) modules/trade/actions/helpers/short-sell.js should handle null inputs:
     AssertionError: the callback wasn't called once as expected
      at Context.<anonymous> (short-sell-test.js:93:3)

  2) modules/trade/actions/helpers/short-sell.js should handle undefined inputs:
     AssertionError: the callback wasn't called once as expected
      at Context.<anonymous> (short-sell-test.js:107:3)
```

tinybike:
@JohnDanz The tests still don't pass, with the same errors as before:
```
  1) modules/trade/actions/helpers/short-sell.js should handle null inputs:
     AssertionError: the callback wasn't called once as expected
      at Context.<anonymous> (short-sell-test.js:94:3)

  2) modules/trade/actions/helpers/short-sell.js should handle undefined inputs:
     AssertionError: the callback wasn't called once as expected
      at Context.<anonymous> (short-sell-test.js:108:3)
```
I'm running these in npm 3.10.6 and Node 6.2.2.

tinybike:
@JohnDanz I did a bit of troubleshooting.  It turns out that the assertion failures I observed before only happen when AURC is npm linked to augur.  (If I clear node_modules and reinstall without npm linking, the tests complete successfully.)  Can you try npm linking with AURC and see if you can reproduce this error?  The tests do need to work with npm link!

PR #800, opened 2016-10-31
Feedback button: Doorbell to UserVoice

tomkysar:
UserVoice can be used for collecting product feedback, and relaying messages to the team. Joey mentioned that we should also put it within the app. I sent UV invites to Jack & Joey, and can add whoever else needs access. Any suggestions written in the box will be posted on the public http://augur.uservoice.com, and any messages sent will hit our UV ticketing system. 

It's a tab anchored to the left vertical of the page right now. Docs for the button are [here](https://developer.uservoice.com/docs/widgets/methods/) and [here](https://developer.uservoice.com/docs/widgets/options/).


tinybike:
@tomkysar: I think we should re-position the Feedback tab (maybe in the bottom left, offset 5% left the way the Doorbell.io tab is now?) -- it looks a bit strange overlaid on the sidebar where it is now.

build/index.html
```diff
@@ -52,8 +52,7 @@
...
-</style>
+	<style type="text/css">#app,#splash,body,html{width:100%;height:100%}#splash,#splash-text{top:0;left:0;overflow:hidden}#splash{display:block;position:relative;color:#f5f5f5;background:#efefef}#splash-text{display:inline-block;position:absolute;bottom:0;right:0;width:50px;height:50px;margin:auto;color:#fbfbfb;background-image:url(2.0.0/assets/images/augur_logo_200_white.png);background-repeat:no-repeat;background-position:center center;background-size:contain;font-family:Impact,"Arial Black","Avant Garde",Charcoal,sans-serif;font-size:20vh;font-weight:700;text-shadow:0 0 50px #e9e9e9;text-align:center;animation-name:splash-anim;animation-duration:10s;animation-iteration-count:infinite;animation-timing-function:cubic-bezier(.11,.63,.25,.89);animation-direction:alternate}@keyframes splash-anim{0%{transform:scale(1) rotate(0)}100%{transform:scale(4) rotate(3deg)}}</style>
```
tinybike: Inline CSS is something we're trying to stay away from.  Styling should all go in AURC.  (Even after the repos are merged, we still don't want inline CSS.)

tomkysar:
@tinybike - updated the styling and positioning to that of the doorbell button.

tinybike:
@tomkysar: the feedback button unfortunately isn't visible once the page has started to load.  Possibly the z-axis setting needs to be increased?

tomkysar:
@tinybike I updated it with z-axis, visibility, and a few more tags from #doorbell-button above. I can't replicate the button going missing once the page has started to load, but hoping this solves the issue.

tinybike:
@tomkysar: I just tested this in both Chrome and FF; it doesn't work in either (the feedback button is not visible).  What browser are you using where this works?

tomkysar:
@tinybike I accidentally removed a div in src/index before that last commit. How about now? I just tried it on Chrome, FF and Safari.

tinybike:
Looks good!  Merging.

PR #805, opened 2016-11-08
1262 report panel

stephensprinkle:


tinybike:
@stephensprinkle The report panel now appears, but when I try to actually submit a report, I get the following error:
```
components.jsx:35371 Uncaught TypeError: this.props.onClickSubmit is not a function(…)
```

tinybike:
Does the new `getValue` function properly iterate over objects with function properties?
```
function getValue(obj, target) {
	return target.split('.').reduce((o, x) => ((typeof o === 'undefined' || o === null) ? o : o[x]), obj);
}
```

tinybike:
Oh -- never mind, I see the problem -- `onClickSubmit={market.onSubmitReport}` should be `onClickSubmit={market.report.onSubmitReport}`.

tinybike:
When the error mentioned above is fixed, the following error takes its place (when submitting a report):
```
components.jsx:1244 Warning: There is an internal error in the React performance measurement code. Did not expect componentDidMount timer to start while render timer is still in progress for another 
components.jsx:34422 Uncaught TypeError: Cannot read property 'orderBook' of undefined(…)
```

PR #810, opened 2016-11-14
1254 changes and fixes

stephensprinkle:
Encompassed within the changes + fixes task, this PR just houses the Augur changes required to fix the filtering issue.


tinybike:
@stephensprinkle: it'd be good to distinguish between "expired" markets (i.e., markets that are past their expiration date but which are still *open* in the sense that they can still be traded on) and "closed" markets (i.e., markets that have been reported on and have had their winning outcomes set).

There's currently a property attached to markets (in the market selector) `isOpen`, which uses the `isMarketDataOpen` method (right above the `isMarketDataExpired` method you edited) to determine whether or not an outcome has been set.  If `market.isOpen` is `false`, then the market is *closed* and can no longer be traded on.

Right now there is a `market.isExpired` property, which is just set to `!market.isOpen`.  I think it would be more useful to set this to `market.isExpired = isMarketDataExpired(marketData, currentTime)` in the market selector.  IMO both "expired" and "closed" should be options in the drop-down.

Is this what was intended with the `reported` option?  Is `market.result` intended to be the same as `market.reportedOutcome` (the outcome of the market once it has been reported on)?  IMO it's easiest to just use the `isMarketDataOpen` method, which checks for this.

PR #818, opened 2016-12-14
Add the packaging metadata to build the augur snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds for adventurous users.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: augur-ui
+version: master
+summary: Augur front-end
+description: |
+  Augur is a decentralized prediction market platform built on Ethereum. This
+  is the reference client. It runs locally in your browser and communicates
+  directly with the ethereum network, without going through intermediate
+  servers.
+
+grade: devel
+confinement: strict
+
+apps:
+  server:
+    command: sh -c \"$SNAP/lib/node_modules/augur-ui && npm start\"
+    plugs: [network, network-bind]
+
+parts:
+  augur:
+    source: .
+    plugin: nodejs
+    npm-run: [build]
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -21,3 +21,4 @@ parts:
     source: .
     plugin: nodejs
     npm-run: [build]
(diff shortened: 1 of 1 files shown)


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/augur
    $ git checkout snapcraft
    $ cd augur
    $ snapcraft
    $ sudo snap install *.snap --dangerous

Open the browser and go to localhost:8080

If you have any questions or comments, please let me know.

elopio:
Nice, thanks @tinybike !
Did you take a look at snapcraft.io ? Now that there's a snap in your project, it would be great if you register it in the store: https://myapps.developer.ubuntu.com/dev/click-apps/register-name/
And after that, we can hook travis to make automatic releases to the edge channel every time there's a change in master, to get feedback from early adopters.

tinybike:
I did skim snapcraft.io.  It looks useful, although of course the devil is in the details.  Am I correct to think of this as a sort of apt-get for dockerized applications, or is that wrong / missing some important nuance?

elopio:
That is actually a good summary in one sentence ^_^

It's like apt-get, but it is also transactional, it has rollbacks, automatic updates and you can release any time to the official archive, you don't have to follow the distro release cycle, and you don't need a distro maintainer to do it for you. Also your snap is self-contained, so you don't have to follow any rules for packaging. Whatever works for you is ok, because your snap can't affect the system and can't affect other snaps.

It's like docker for applications, but it's more lightweight because it runs directly on top of the operating system, you don't need an additional OS to isolate your app. And you can share any resources you want from the system and other snaps through the secure use of interfaces (the plugs section in this yaml declares the interfaces that your snap will connect to). So a full system of snaps can run in an armhf system like raspberrypi without a lot of overheard.

I would love if you give it a try and let me know your opinions, what you find useful or what you find confusing. We are talking to a few upstreams to gather feedback that will help us set the priorities of our backlog for the next months.

Thank you!

PR #821, opened 2016-12-16
Build the snap in travis

elopio:


elopio:
I'm not sure how will you feel about me touching your travis file. Please take a look and let me know.

tinybike:
I'm not clear why it's preferable to use snap in travis, versus the way it's done now (which works fine, as far as I know).  Can you explain?

elopio:
Hey @tinybike. This is just one more test added to the travis build matrix, to make sure that the changes in a pull request still result in a buildable snap. The other existent tests will still run as before without changes.

So this will ensure that your master is always in a releasable state, enabling the next step that's to actually deploy to the edge channel in the store every time there's a change in master.

tinybike:
@elopio while I'm happy to have the config file in our repo (so people can use Snapcraft if they want to), but I'm going to hold off on making this part of our travis build matrix for now -- it's one more thing that can break, and our builds have an awful lot of moving parts already!

elopio:
@tinybike that's understandable. I will be available always in case you need a hand, but of course I know that it could cause delays. I think you will like what we are cooking for early next month. It is a github integration that will build your snaps for all the architectures without using travis as an intermediary. I will ping you again when it's ready for you to try. In the meantime, let me encourage you to register augur in the store and push an initial snap to the edge channel. That way you can start gathering feedback and usage stats, and the vast community of ubuntu users can start doing some early testing. I think that will help you deciding if it is worth for your project to maintain the snap.

pura vida.

elopio:
hey @tinybike, I'm back at work and catching up with my pending conversations from last year.
Did you have any chance to take a look at the Ubuntu store?

Happy new year!

PR #833, opened 2016-12-23
1426 reporting tab

stephensprinkle:


tinybike:
I don't see the reporting tab -- does it appear for all markets, or only markets that are reported on already?

tinybike:
Reading thru the code.  Ok, this makes it so the trade stuff is still accessible when reporting, makes sense -- merging.  Might be good to make the reporting tab always visible (even when not up for reporting), but can make a new task for that later!

PR #839, opened 2017-01-04
1505-text-selection

stephensprinkle:


src/modules/app/components/app.jsx
```diff
@@ -82,6 +82,8 @@ export default class App extends Component {
 
 	// Touch Events
 	attachTouchHandler() {
+		delete Hammer.defaults.cssProps.userSelect; // Allows for text selection
```
tinybike: Hammer disables text selection by *default*?  Sheesh...

PR #845, opened 2017-01-09
#1240 - Added trailing ellipsis for market outcome previews

tomkysar:
Market outcome name previews will now trail with (...) depending on screen width. I'm not sure why, but the width needs to be set to prevent that right column from expanding its width endlessly on a long name. It scales to always one line.

Example:

screen shot 2017-01-09 at 12 34 01 pm

### DIFF
--- a/src/modules/market/less/market-preview-outcomes.less
+++ b/src/modules/market/less/market-preview-outcomes.less
@@ -22,7 +22,11 @@
 
 		.outcome-name {
 			flex: 3;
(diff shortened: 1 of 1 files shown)


tinybike:
Could you add a popup with the full outcome text?  Otherwise this looks good!

tinybike:
@tomkysar I'm going to go ahead and merge this PR.  I created a new CH issue for the popup!

PR #851, opened 2017-01-19
#1874 - Added popover for long outcome names

tomkysar:
Added a popup with the full outcome name preview for outcomes that are longer than 35 characters (appx. breakpoint of full screen where truncation starts).


tinybike:
@tomkysar This isn't working quite right.  In markets with more than one long outcome description, the popover only displays for the first one.

PR #856, opened 2017-02-07
1646-trade-lock

stephensprinkle:


tinybike:
One thing that would be good to add is to allow making new orders (bid/ask) but not taking orders (trade).  Will put this in a separate issue.

PR #863, opened 2017-02-20
2142 topic markets

stephensprinkle:


src/modules/market/selectors/market.js
```diff
@@ -291,7 +291,7 @@ export function assembleMarket(
 
       market.tags = (market.tags || []).map((tag) => {
         const obj = {
-          name: tag && tag.toString().toLowerCase().trim(),
+          name: tag && tag.toString().trim(),
```
tinybike: @stephensprinkle Dumb as it sounds, I don't think we even want to trim the tags, since trimmed tags will not match untrimmed tags on-contract.  I can't think of a valid use-case for untrimmed tags, so I'm making it so `formatTag` in augur.js trims tags (during `createMarket`).  However it's always possible some enthusiast will send an `     untrimmed tag rofl   ` to the Serpent API directly and that probably should cause his market to be completely invisible to the UI!

tinybike: Removing the trims and committing, then merging, rest of this looks good!

stephensprinkle: Sounds good!  

Yeah, it shouldn’t impact the ultimate display of the market; however, would we still want to trim w/in the components (All Tags, Tags in market preview, etc.), so we don’t have odd leading/trailing whitespace?

tinybike: My feeling is that we should display the tags as-is, even if they include stupid leading/trailing whitespace.  Presumably people will refrain from doing that b/c it will make their tag unique and dumb, and very far down the popularity list!

stephensprinkle: 👍

PR #864, opened 2017-02-20
2151-bug-url-redirect

stephensprinkle:
This bug was due to the method `displayLoginMessageOrTopics` always redirecting a user to the topics view, even when a specific path was present.


tinybike:
Good catch.  I noticed this bug yesterday and was wondering what was causing it!

PR #878, opened 2017-03-12
Updates to handle node fallback.

MicahZoltu:
Previously, ethrpc was doing hosted node fallback.  This code has been moved up to the UI instead, so everything else just operates on an array of possible nodes, the UI provides the list of possible nodes to connect to.

Also adds a launch configuration for debugging tests in VSCode and fixes a minor issue in building on Windows.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,37 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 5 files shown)


MicahZoltu:
Note: This should not be merged until ethrpc, ethereumjs-connect, and agurjs PRs are merged.

Issue #15, opened 2015-04-26
Holdings component


niran:
From https://github.com/AugurProject/augur-client/wiki/Specification:

The holdings page displays a list of event outcomes with the number of outcome tokens held, their current price, and the current total value of the user's holdings of a given outcome. It also displays the total value of all held event outcomes. Each outcome links to its corresponding event detail page, where the user can buy and sell shares.

Users can send reputation to other accounts from the holdings page.

scottzer0:
shifting this to be move of an account overview (asset balances/actions, recent transaction, upcoming votes, etc.) with options to dive deeper when appropriate (see all your transaction, etc.)

niran:
I think upcoming votes should go in the ballots component. Most users are traders or observers, who don't have any reputation to vote with. They should be able to ignore the ballots component altogether and never come across voting information that is irrelevant or confusing to them.

The main goal of the holdings component should be to tell people about their funds that are tied up in Augur. Transactions per asset and overall would probably be better off as a page you can get to from the holdings screen. Each asset could have a "transaction history" link.

PR #880, opened 2017-03-16
1340 create market reskin

stephensprinkle:


tinybike:
I love the new look and feel!  It is a huge improvement over the old create market process.  A few issues that should be fixed prior to merging:

- Clickable things (market type button, the back/next buttons, etc.) should have cursor: pointer
- Clicking on "clear new market" takes me to a state where I can't progress any further and there are not back or next buttons etc:

screenshot_2017-03-16_17-12-54
- Clicking on "back" also takes me to this state
- For local time display, should explicitly show the time zone being used
- Adding orders to the order book displays this warning in the console:
```
warning.js?8a56:36 Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `CreateMarketPreview`. See https://fb.me/react-warning-keys for more information.
    in li (created by CreateMarketPreview)
    in CreateMarketPreview (created by CreateMarketView)
    in div (created by CreateMarketView)
    in section (created by CreateMarketView)
    in CreateMarketView (created by Connect(CreateMarketView))
    in Connect(CreateMarketView)
    in Routes (created by App)
    in div (created by App)
    in div (created by App)
    in div (created by App)
    in div (created by App)
    in main (created by App)
    in App
    in AppContainer
    in Provider
```
- My best bid is at 0.5.  When I try to place a best ask at 0.6, it throws an error when I've typed "0." because it thinks that is lower than 0.5.  Suggest only doing this check after the full value is entered.
- Repeated instances of `Highcharts error #15: www.highcharts.com/errors/15` when adding or removing orders
- The market link from the transactions display to the market detail page does not work after creating the market
- Min/max outcome display when creating a scalar market looks incorrect, should probably show the max and min values in the upper right:

screenshot_2017-03-16_17-22-42
- Binary markets probably should not show the outcome at all in the upper right (since it's just "Yes") or in the order book creation display:

screenshot_2017-03-16_17-16-41
- Scalar markets also probably should not show the outcome in the order book creation display, since this looks a little silly:

screenshot_2017-03-16_17-26-20

stephensprinkle:
Sounds good — I’ll work to get those wrapped up.

I also woke up in the night and realized I’d failed to switch over the chart to bar, so I’ll do that as well.

stephensprinkle:
All items highlighted above have now been addressed.  

A few points:

> The chart looked odd w/ bars, so I updated the chart to utilized ‘area’ instead.  Effectively a line chart w/ the bounded area shaded, effectively resulting in the UX that I believe was intended w/ a bar chart.

> The transactions display for create market will only link to the market once we’ve received back a market id.  I double checked this and it is the current exhibited behavior.  

> Re: order entry — those inputs currently have a debounce of 500ms.  This is mostly there to allow for the inference that the user is ‘done’ inputing their value and should thus have it validated.  The other approach is to validate on blur, but this probably wouldn’t be a great UX for this form.

tinybike:
- When I go to the order creation panel, then click "back" (without adding any orders), the "next" button says "review" and takes me directly to the review page, instead of to the order book page.
- When adding orders, I received the following error in the console:
```
warning.js?8a56:36 Warning: flattenChildren(...): Encountered two children with the same key, `0:$Issufficient funds`. Child keys must be unique; when two children share a key, only the first child will be used.
    in ul (created by CreateMarketFormInputNotifications)
    in CreateMarketFormInputNotifications (created by CreateMarketFormOrderBook)
    in div (created by CreateMarketFormOrderBook)
    in div (created by CreateMarketFormOrderBook)
    in div (created by CreateMarketFormOrderBook)
    in form (created by CreateMarketFormOrderBook)
    in div (created by CreateMarketFormOrderBook)
    in div (created by CreateMarketFormOrderBook)
    in article (created by CreateMarketFormOrderBook)
    in CreateMarketFormOrderBook (created by CreateMarketForm)
    in article (created by CreateMarketForm)
    in CreateMarketForm (created by CreateMarketView)
    in div (created by CreateMarketView)
    in section (created by CreateMarketView)
    in CreateMarketView (created by Connect(CreateMarketView))
    in Connect(CreateMarketView)
    in Routes (created by App)
    in div (created by App)
    in div (created by App)
    in div (created by App)
    in div (created by App)
    in main (created by App)
    in App
    in AppContainer
    in Provider
```
Two issues here: 1) this error didn't propagate to the UI at all (not sure if it's supposed to, though), and 2) my account has much more than sufficient funds for the order -- I think the UI may be checking the amount of "real ETH" the account has rather than the amount of "ETH" (which is the relevant quantity).
- The depth chart on the review panel displays like the old chart (i.e., unfilled).

tinybike:
- I receive the following error in the browser console while zooming in and out:
```
Uncaught TypeError: Cannot read property 'offsetHeight' of null
    at Footer.handleWindowResize (eval at <anonymous> (main.ca006fb….js:5540), <anonymous>:71:22)
    at Footer.handleWindowResize (eval at <anonymous> (main.ca006fb….js:8007), <anonymous>:44:30)
    at later (eval at <anonymous> (main.ca006fb….js:2173), <anonymous>:17:12)
```
- When I try to add orders to a categorical market, I get the following error in the console and the order does not add to the book:
```
Uncaught Error: gt() not a number: undefined
    at raise (eval at <anonymous> (main.ca006fb….js:432), <anonymous>:1190:25)
    at eval (eval at <anonymous> (main.ca006fb….js:432), <anonymous>:1178:33)
    at new BigNumber (eval at <anonymous> (main.ca006fb….js:432), <anonymous>:193:67)
    at BigNumber.P.greaterThan.P.gt (eval at <anonymous> (main.ca006fb….js:432), <anonymous>:1503:35)
    at CreateMarketFormOrderBook.validateForm (eval at 1566 (2.e506154….js:166), <anonymous>:429:33)
    at CreateMarketFormOrderBook.validateForm (eval at <anonymous> (main.ca006fb….js:8007), <anonymous>:44:30)
    at Object.onChange (eval at 1566 (2.e506154….js:166), <anonymous>:613:41)
    at eval (eval at 1502 (4.e5ed6ab….js:15), <anonymous>:98:35)
```
I tried both outcome 0 and 1.  The values I'm entering should be valid for this market:

screenshot_2017-03-18_11-00-37

tinybike:
- It would be a really nice little feature if we could tab from the amount to the price field when creating the order book, and press "Enter" to add the order -- this will be quite repetitive as it is, so little things like this to speed it up will probably be appreciated by users.
- Can we change the scalar outcome display to explicitly denote the min and max values displayed as min and max values?  I think it's a bit confusing at the moment:

screenshot_2017-03-18_11-03-09

tinybike:
Hmm, the links to the new market from the transactions page still aren't working for me.

tinybike:
Update: now it seems to be working fine.  Hm.  Maybe I had not cleared my cache?

stephensprinkle:
@tinybike — Gremlins :). Possibly, but not sure. I didn’t refactor transactions at all as part of this PR and the last touch point related to the market ID w/ transactions was a few PRs back I believe.

tinybike:
Agree, gremlins seems like the most likely explanation to me too

PR #882, opened 2017-03-27
655 positions reskin

stephensprinkle:


tinybike:
Hmm, positions aren't updating dynamically -- I have to refresh to see the positions listed.

stephensprinkle:
K, let me take a look.

tinybike:
Unethical thumbs-down should be red!
screenshot_2017-03-28_12-15-14

stephensprinkle:
Ah, misspelled that class!  Just pushed a fix.

stephensprinkle:
Hang on, this change didn’t quite resolve things well.

stephensprinkle:
OK, the fix looks to be working properly.  Let me just update the tests + the PR should be GTG again.

tinybike:
Not 100% sure if this is due to a change on this branch, but the transactions history is now loading in an infinite loop (and pegging my CPU)...

tinybike:
Update: it appears to also be infinitely reloading the order book for one particular market (0xdf4e99d24efb76b9d0b7ea389b8add7ddf260242f39407eac15638969224b1d5).

tinybike:
Update x2: I can't reliably reproduce the infinite reload, and am not sure it's caused by this PR, so going to go ahead and merge this.  Looking good btw! :grin:

PR #884, opened 2017-03-31
2349 reselect

tinybike:


src/modules/account/selectors/core-stats.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export const selectOutcomeLastPrice = (marketOutcomeData, outcomeID) => {
+  if (!marketOutcomeData || !outcomeID) return null;
+  return (marketOutcomeData[outcomeID] || {}).price;
+};
+
+// Period is in days
+export const createPeriodPLSelector = period => createSelector(
+  selectAccountTradesState,
```
stephensprinkle: Per the docs, shouldn’t there be just two args passed to the `createSelector`: `an array of input-selectors and a transform function as its arguments`?

src/modules/app/components/routes.jsx
```diff
@@ -57,15 +57,12 @@ export default class Routes extends Component {
...
-          account: p.loginAccount,
-          onChangePass: p.loginAccount.onChangePass,
-          authLink: (p.links && p.links.authLink) || null,
-          onAirbitzManageAccount: p.loginAccount.onAirbitzManageAccount
-        };
-        import('modules/account/components/account-view').then((module) => {
+        import('modules/account/container').then((module) => {
           const AccountView = module.default;
+          viewProps = {
+            loginMessageLink: p.links.loginMessageLink,
```
stephensprinkle: This might not have been addressed yet due to being in-process, but we may want to move all values passes via props to the container as well.

src/modules/markets/selectors/tags.js
```diff
@@ -1,46 +1,49 @@
-import memoizerific from 'memoizerific';
+import { createSelector } from 'reselect';
+import store from 'src/store';
+import { selectSelectedTagsState } from 'src/select-state';
+import { selectFilteredMarkets } from '../../markets/selectors/markets-filtered';
```
stephensprinkle: As I’ve been re-skinning, I’ve been updating all paths to utilize the path aliases as outlined w/in the `.babelrc` file.

tinybike:
You can pass inputs to createSelector either as an array or as separate
parameters... Separate params seemed cleaner to me but either way works I
think.

tinybike:
Totally agree re: value passes.  IMO only "reasonably complicated" derived
data is worth putting in a selector/memoizing.  Anything else, the gain in
simplicity by putting it in the container is better!

src/select-state.js
```diff
@@ -0,0 +1,69 @@
+import { createSelector } from 'reselect';
```
stephensprinkle: Would we want to place these input selectors contextually within their respective selector files as opposed to having them all within a single file?

My other assuming in reviewing these changes is that these changes will supersede the existing selector approach of having a global selector object, correct?

src/modules/markets/selectors/tags.js
tinybike: Does proxyquire use the path aliases as well?  Generally I've had to keep things using relative paths if I need to stub them with proxyquire -- maybe I am not specifying the paths correctly?

src/select-state.js
tinybike: > Would we want to place these input selectors contextually within their respective selector files as opposed to having them all within a single file?

Yes, we probably should!  I put all the selectors that are nothing more than pulling a single prop out of state in a single file for convenience while developing, but we probably should disperse all these "selectors" into their respective modules.

> Another assuming I had in reviewing was that these changes will supersede the existing selector approach of having a global selector object, correct?

Yep!  The global selector object is still there since the conversion isn't 100% complete yet, but my goal here is to get rid of the "selector object" entirely and replace it with selector functions -- since a selector is, after all, by definition a function and (to me) it's weird and confusing to have "selectors" which are just sitting in a global object and looking for all the world like a second state object.

src/modules/markets/selectors/tags.js
stephensprinkle: Currently, no, but I should get that setup.  

Right now if you have utilized a path alias, you’ll need to use the shortest path to the referenced file in order for proxyquire to work.

src/select-state.js
stephensprinkle: Ok, awesome, that all sounds good!

PR #885, opened 2017-04-01
Upgrade the Airbitz SDK to edge login v2

swansontec:
This pull request upgrades to the latest version of the Airbitz SDK.

We have re-written edge login from the ground up, which makes this deployment a little tricky. We plan to release the new version of our mobile app on Monday afternoon, and upgrade our production servers at the same time. At that point, edge login v1 will stop working and these changes will be necessary.

Until Monday comes around, our infrastructure is still on edge login v1. To test this pull request, you will need to use the RC version of our mobile app (https://developer.airbitz.co/augur-recovery/airbitz-develop-release-2017032904.apk), as well as the test version of our auth server. This can be accomplished by running the following code in the browser terminal:

    localStorage.setItem('airbitzAuthServer', 'https://test-auth.airbitz.co/api')

Obviously, all this complexity will all go away on Monday, when we release the latest version of our app and upgrade our servers to match.

We have also created a key recovery tool for extracting keys from the v1 login system, in case anybody sends money to them. You can see that here: 

https://developer.airbitz.co/augur-recovery/

Let me know if you need anything else.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "link": "npm link augur.js && npm link keythereum"
   },
   "dependencies": {
-    "airbitz-core-js-ui": "0.0.14",
+    "airbitz-core-js-ui": "0.1.2",
     "async": "2.1.5",
     "augur.js": "3.15.3",
     "bignumber.js": "4.0.0",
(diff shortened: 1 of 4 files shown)


swansontec:
I have upgraded our core to 0.1.3, which should hopefully address the broken unit test from earlier.

PR #895, opened 2017-05-19
Reduce sidebar width to 260px and truncate tags

JohnAllen:


JohnAllen:
This doesn't handle hovering/displaying the entire tag as @tinybike suggested in CH.  Happy to add something for that if there are suggestions as to correct approach.  

I don't think we want `<ReactTooltips />` on each one because then each one will pop something up which IMO would be obtrusive (assuming the same behavior as on that packages demo site).  But I don't know, maybe a conditional `<ReactTooltip /> if (s.length > x)`?  Let me know thoughts.

stephensprinkle:
I’m going to close this PR in favor of a PR from a branch in this repo (makes things easier to make changes).

PR #896, opened 2017-05-19
Sidebar width

stephensprinkle:


stephensprinkle:
A few changes I made:

Since the styling of the ‘checkboxes’ is specific to it’s utilization w/in the side-bar, I moved those styles so their more narrowly scoped.

I updated the layout so that flex will establish the elements respective widths — this has the benefit of letting the truncation be fully responsive to both parent width constraints + label width constraints.  As the amount of markets increases, the label will be truncated accordingly. 

As to your original question:

I’d agree that we should only do a tooltip for truncated labels.  

One way to do this would be to set a ref on each tag and inspect the rendered text and compare that against the full string, if it’s not equal, then append the tooltip for that tag.

JohnAllen:
For the refs this stateless component is going to need to be changed to a proper React class, correct because it will need a method to append the tooltip, correct?

JohnAllen:
There's probably not a way to do this in the current stateless function form, right?  

Also, there's a `<ReactTooltip />` already in this component as you probably know.  And the `data-tip` is fed a prop that isn't passed for this instance of `checkbox` (`{p.title}`).  (Wondering if `<Checkbox />` works elsewhere for other instances or if it needs to be modified)

stephensprinkle:
That’s correct, it would need to become a stateful component (which is fine).  

The `title` prop is optional (for the reason you mentioned re: optional tooltip), so it’s totally valid to exclude it from the passed props.  

AFAIK, everything is fine w/ Checkbox inclusions elsewhere.  Those occurrences may be going away however in light of the reporting changes…TBD.

Side Note: We’re not 100% on the prob validation coverage yet (a TODO before release), but generally referencing the validations will give you an idea of what is optional and what’s not.

stephensprinkle:
Closing this PR for now until it’s ready for re-submission.

PR #898, opened 2017-05-31
Display potential profit and loss to user as they enter their order

JohnAllen:


src/utils/calc-order-profit-loss-percents.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param maxValue only relevant for scalar markets; all other markets max is created and set to 1
+ * @returns object with the following properties
+ *    potentialEthLoss:       number, maximum number of ether that can be lost according to the current numShares and limit price
+ *    potentialEthProfit:     number, maximum number of ether that can be made according to the current numShares and limit price
+ *    potentialProfitPercent: number, the maximum percentage profit that can be earned with current numShares and limit price,
+ *                                    excluding first 100% (so a 2x is a 100% return and not a 200% return). For BUYs, loss is always 100% (exc. fees)
+ *    potentialLossPercent:   number, the max percentage loss that can be lost with current numShares and limit price; for SELLs loss is always 100%
+ */
+
+export default function (numShares, limitPrice, side, minValue, maxValue, type) {
```
dandv: Given this has 6 parameters, better make it an options object?

test/utils/calc-order-profits-loss-test.js
```diff
@@ -0,0 +1,439 @@
...
+      actual: {
+        potentialEthLoss: 597.05,
+        potentialEthProfit: 17.255,
+        potentialProfitPercent: 100,
+        potentialLossPercent: 488.23
+      }
+    }*/
+  ];
+
+  orders.forEach((order) => {
```
dandv: `for (const order of orders) { ... }`

stephensprinkle: Great point!  Yeah, let’s go ahead and make this change @JohnAllen.

src/utils/calc-order-profit-loss-percents.js
stephensprinkle: Great suggestion!  While I do generally prefer an options object, for consistency in the UI CB, let’s leave the multiple params in place here for now.

stephensprinkle:
@JohnAllen — A few major things I changed:

- I moved the guarding conditional into the method itself.
- I removed the calls for `toString` as the formatting of the numbers can handle BN
- With that change in place I could then used the calculations from the PL for eth to simplify the calculations for percent.
- Added the bottom border to the UI.  
- Updated the tests to be more explicit/narrowly scoped.  

Let me know if you’ve got any thoughts/questions re: the changes!

I’ve got one more test to update, but after that I’ll be merging the PR.

PR #900, opened 2017-06-06
Change classnames module imports and usage to classNames

JohnAllen:
Pretty self-explanatory

I also removed those deeply-nested object props in outcome-trade-summary.jsx


stephensprinkle:
Looks good!

PR #903, opened 2017-06-22
660 account reskin

stephensprinkle:


tinybike:
On the Transfer page, the text `Use this form to send your ETH Tokens, ETH, and REP to another address.` should be updated to `Use this form to send your ETH or REP to another address.`

tinybike:
Also on the Transfer page, `Tranfer Funds` is misspelled (should be `Transfer Funds`).

tinybike:
Should probably remove the "Download Login ID" button from the export page, since the whole concept of login ID is about to be deprecated + the point of login ID was to have a string instead of a file.

tinybike:
On the Export page, `Use either the QR code or download link save a copy of the key file for import into a wallet.` should have the word `to` between link and save

tinybike:
For some reason, entering an account name on the account page causes Chrome to max out my CPU as I'm typing.  Is something being updated behind the scenes every key press?

tinybike:
The Ethereum address of the account (below the account name, on the account page) should always display the entire address.  I can't think of a situation where someone would want to see `0x94...b2c1` instead of their actual address.

tinybike:
Addresses should be displayed (and verified) using [EIP55](https://github.com/ethereum/EIPs/issues/55) (capitals-based) checksums, the same way that Mist, MEW, etc. do it.

stephensprinkle:
- [ ] Update Copy on Transfer View
- [ ] Fix typo on Transfer View
- [ ] Remove Download LoginID
- [ ] Update copy on export view
- [ ] Check into account name change CPU utilization
- [ ] Always Display the entire address in account details component

stephensprinkle:
Ok, all requested changes are now in place.  

Re: the name change — I’ve moved that to be housed in local storage.  The slowdown was that since it was stored w/in the loginID, we were re-encoded on change.

stephensprinkle:
Btw — looks like somehow my sig changed…I am signing on my end, let me update things in my account.

stephensprinkle:
OK, signature issue fixed — ended up being a mis-configuration on my end related to `user.email`.

Mahmoudhpp (approved):

PR #908, opened 2017-06-29
fix CreateMarketPreview setState error

JohnDanz:
Because of the way setState was being called, sometimes it would get triggered after createMarketPreview had been unmounted which resulted in an error. Attempting to set the state of an unmounted component. This now checks to make sure the component is mounted, also corrects a similar situation that was more rare where a style was being applied to an unmounted component because of a timeout.


JohnAllen:
Glad we reached the same conclusion on how to fix this one and you beat me to it!

PR #909, opened 2017-06-29
2668-TransactionLabelsToConstants

JohnDanz:
Story 2668: https://app.clubhouse.io/augur/story/2668/update-transaction-labels-to-be-constants

modified the `SHORT_ASK` transaction type constant to be 'shortAsk' instead of 'short_ask' to match the api method. added constants instead of hardcoded values for transaction actions. added `TRADE` (`'trade'`), `COMMIT_TRADE`(`'commitTrade'`), and `CANCEL`(`'cancel'`) to transaction type constants.


stephensprinkle:
Leaving open for anticipated changes discussed in Slack.

PR #911, opened 2017-07-06
267 export transaction history

JohnDanz:
CH Story 267: https://app.clubhouse.io/augur/story/267/export-of-transactions-history

Adds an Export Transaction button to the load-transactions-actions component which is used on the transactions, my-markets, my-reports, and my-positions pages.


stephensprinkle:
A few things to resolve:

- Per the task spec: 
 - `button should be right of the ‘load more’ and ‘load all’ buttons with margin between them.`
  - Wasn’t specified, but 1em is good here
 - `Button copy: ‘Export All’`
- When the button is disabled, the cursor should be default instead of pointer.
- By convention, add `-test` to the end of the filename for the new test file.
- While not originally part of the spec, exclude export button from the portfolio views.

I made a couple other small adjustments to `load-account-history` and actually there are a few areas of cleanup we can do, but that’s out of scope for this task.

stephensprinkle:
Functionally looks good!  I’ll just wait on the test updates, then I’ll merge on down.

stephensprinkle:
Added an additional test for the `loadMoreTransactions` method.  Merging down once things pass.

PR #919, opened 2017-08-16
Animated graph background integration.

smallfx:
Hardware accelerated + responsive. 📈✨

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "keythereum": "1.0.0",
     "memoizee": "0.4.5",
     "moment": "2.18.1",
+    "p5": "^0.5.11",
     "prerender-node": "2.7.1",
     "prop-types": "^15.5.10",
     "qrcode.react": "0.7.1",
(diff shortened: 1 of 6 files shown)


stephensprinkle:
In further testing, it’s looks to be related to the window size as exhibited here: https://www.youtube.com/watch?v=PXGx53aFfOk

stephensprinkle:
NOTE — just waiting on the p5 warning messages resolution, then this should be good to merge.

stephensprinkle:
@smallfx — I’m going ahead and closing this for the time being.

smallfx:
So, modifying the P5 library didn't work as I expected after research... though having read all of the renderer code, I figured out that it had some slightly unexpected requirements when in "immediate mode" (just drawing primitives directly). Seems like line shapes need `.fill()` set for every `.shapeBegin()`, which I don't even think they use (should be using `.stroke()` color). Anyway, setting it for each shape seems to have no detrimental effect on performance.

stephensprinkle:
🎉

stephensprinkle:
Thanks for the extra detail!

PR #920, opened 2017-08-19
Simple prop validation fix

smallfx:
`selectedCategory` prop is not actually required by `InnerNav` - it defaults to `null` in the Redux state, which is expected and accounted for.


stephensprinkle:
👍

PR #921, opened 2017-08-20
2744 routing refactor

tinybike:
Creating a PR for more convenient commenting / discussion.


package.json
```diff
@@ -44,10 +46,13 @@
     "react-blockies": "1.1.5",
     "react-datetime": "2.8.10",
     "react-dom": "15.6.1",
+    "react-helmet": "5.1.3",
     "react-highcharts": "12.0.0",
     "react-hot-loader": "3.0.0-beta.7",
     "react-modal-dialog": "4.0.7",
     "react-redux": "5.0.5",
+    "react-router-dom": "4.1.1",
+    "react-router-redux": "next",
```
tinybike: Recommend using a fixed version number for all dependencies.

package.json
```diff
@@ -33,7 +33,9 @@
     "hammerjs": "2.0.8",
     "helmet": "3.6.1",
     "highcharts": "5.0.12",
+    "history": "4.6.3",
     "keythereum": "1.0.0",
+    "lodash": "^4.17.4",
```
tinybike: Two comments:
1. We have a long-standing rule that "kitchen sink" libraries like lodash are not permitted.  Can you import whatever methods you need in isolation, rather than the entire library?
2. Should use a fixed version number for all dependencies.

package.json
```diff
@@ -33,7 +33,9 @@
     "hammerjs": "2.0.8",
     "helmet": "3.6.1",
     "highcharts": "5.0.12",
+    "history": "4.6.3",
```
tinybike: Why is this dependency necessary?

src/history.js
```diff
@@ -0,0 +1,3 @@
+import createHistory from 'history/createHashHistory';
+
+export default createHistory();
```
tinybike: This file would benefit from a comment explaining its purpose.  It's not clear to me what the `history` module is being used for here / why we need it.

src/main.jsx
```diff
@@ -30,26 +34,22 @@ if (process.env.NODE_ENV === 'development') {
...
+          <Routes />
+        </HashRouter>
       </AppContainer>
     </Provider>,
     document.getElementById('app')
   );
 }
 
-store.subscribe(handleRender);
+handleRender();
```
tinybike: Where does React subscribe to the Redux store / why was this moved?

src/main.jsx
```diff
@@ -30,26 +34,22 @@ if (process.env.NODE_ENV === 'development') {
...
 
-store.dispatch(updateURL(window.location.pathname + window.location.search));
-
-window.onpopstate = (e) => {
-  store.dispatch(updateURL(window.location.pathname + window.location.search));
-};
-
 store.dispatch(initAugur());
 
 function render(App) {
```
tinybike: `App` parameter is no longer used.

src/modules/app/components/routes.jsx
```diff
@@ -1,94 +1,72 @@
...
-      PropTypes.number,
-      PropTypes.string
-    ]).isRequired,
-    setSidebarAllowed: PropTypes.func.isRequired
-  }
+import App from 'modules/app/container';
 
-  constructor(props) {
-    super(props);
+import asyncComponent from 'modules/common/components/async-component';
```
tinybike: The purpose of `asyncComponent` appears to be to abstract away boilerplate present in the old `routes.jsx`.  While it does make the callsites in this file a bit cleaner, I think on balance introducing a new component class for this is overkill / is a net complexity increase.

src/modules/app/components/side-bar.jsx
```diff
@@ -1,41 +1,167 @@
...
-            text2={`(${tag.numMatched})`}
-            isChecked={tag.isSelected}
-            onClick={tag.onClick}
-          />
-        ) : <NullStateMessage message={nullMessage} />
+import parseQuery from 'modules/app/helpers/parse-query';
+import parseStringToArray from 'modules/app/helpers/parse-string-to-array';
+import makeQuery from 'modules/app/helpers/make-query';
+
+import orderBy from 'lodash/orderBy';
```
tinybike: Recommend moving library imports above imports from our modules.

src/modules/app/helpers/parse-query.js
```diff
@@ -0,0 +1,21 @@
+/**
+  * Parses the raw query string into a formatted array
+  * @param {string} searchString - The raw query param string
+  * @returns {object}
+*/
+export default function parseSearch(searchString) {
```
tinybike: Should probably rename this function to `parseQuery` to match its name when imported (and to match the name of the file).

src/modules/auth/actions/load-account-data.js
```diff
@@ -4,13 +4,11 @@ import { loadAccountDataFromLocalStorage } from 'modules/auth/actions/load-accou
 import { loadRegisterBlockNumber } from 'modules/auth/actions/load-register-block-number';
 import { updateAssets } from 'modules/auth/actions/update-assets';
 import { updateLoginAccount } from 'modules/auth/actions/update-login-account';
-import { displayTopicsPage } from 'modules/link/actions/display-topics-page';
 
 export const loadAccountData = (account, redirect) => (dispatch, getState) => {
```
tinybike: The `redirect` parameter is no longer used.

src/modules/auth/actions/update-is-logged.js
```diff
@@ -0,0 +1,10 @@
+export const UPDATE_IS_LOGGED = 'UPDATE_IS_LOGGED';
+
+export const updateIsLogged = isLogged => (
```
tinybike: Suggest renaming this to `updateIsLoggedIn`; I think `isLogged` is a bit ambiguous.

src/modules/auth/components/auth-import.jsx
```diff
@@ -27,6 +30,10 @@ export default class AuthImport extends Component {
     window.addEventListener('error', this.handleRecoverError);
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('error', this.handleRecoverError);
```
tinybike: What is the purpose of this listener?

src/modules/auth/components/auth-view.jsx
```diff
@@ -8,7 +9,7 @@ import AirbitzLogoIcon from 'modules/common/components/airbitz-logo-icon';
 
 import ComponentNav from 'modules/common/components/component-nav';
 
-import { AUTH_SIGNUP, AUTH_LOGIN, AUTH_IMPORT } from 'modules/app/constants/views';
+import { SIGNUP, LOGIN, IMPORT } from 'modules/app/constants/views';
```
tinybike: Nitpick: "sign in/out" and "log in/out" are both used throughout the code.  It would be good to pick one of these and use it.  (The middleware uses "log in/out", FWIW.)

src/modules/auth/components/authenticated-route.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { connect } from 'react-redux';
+import { Route, Redirect } from 'react-router-dom';
+
+import { AUTHENTICATION } from 'modules/app/constants/views';
+
+import makePath from 'modules/app/helpers/make-path';
+
+import getValue from 'utils/get-value';
+
+const AuthenticatedRoute = ({ component: Component, isLogged, ...rest }) => (
```
tinybike: The `component` prop appears to be unused.

src/modules/common/components/async-component.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+
+// { moduleName, loader, props }
+export default options => (
+  class AsyncComponent extends Component {
```
tinybike: See `routes.jsx` for comments on this component.

src/modules/common/components/async-component.jsx
```diff
@@ -0,0 +1,38 @@
...
+      };
+
+      this.loadAsyncComponent = this.loadAsyncComponent.bind(this);
+    }
+
+    componentWillMount() {
+      this.loadAsyncComponent(options.loader, options.props);
+    }
+
+    loadAsyncComponent(loader, props) {
```
tinybike: `props` parameter is unused here.

src/modules/market/actions/load-full-market.js
```diff
@@ -7,12 +7,16 @@ import { addMarketLoading, removeMarketLoading } from 'modules/market/actions/up
 export const loadFullMarket = marketID => (dispatch, getState) => {
   dispatch(addMarketLoading(marketID));
 
+  // TODO -- improve conditional statement, leaving more verbose loading for now
```
tinybike: Why is this commented out?

src/modules/market/components/market-view.jsx
```diff
@@ -1,31 +1,96 @@
...
+    market: PropTypes.object.isRequired
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isAvailable: getValue(props, 'market.id'),
+      isOpen: getValue(props, 'market.isOpen'),
+      marketId: null
```
tinybike: Recommend using either `marketID` or `marketId` consistently.

src/modules/markets/components/markets-list.jsx
```diff
@@ -1,34 +1,114 @@
...
-        return (
-          <MarketPreview
-            key={market.id}
-            loginAccount={p.loginAccount}
-            {...market}
-            selectedShareDenomination={selectedShareDenomination}
-            shareDenominations={shareDenominations}
+import debounce from 'utils/debounce';
+
+export default class MarketsList extends Component {
```
tinybike: General comment: `MarketsList` is notably more complex than before, and it has pagination logic scattered throughout.  This feels "off" to me, although I haven't studied the new pagination code enough to make any specific recommendations.

src/modules/markets/components/markets-view.jsx
```diff
@@ -1,43 +1,156 @@
...
+import parseQuery from 'modules/app/helpers/parse-query';
+import isEqual from 'lodash/isEqual';
+
+import parsePath from 'modules/app/helpers/parse-path';
+import makePath from 'modules/app/helpers/make-path';
+
+import { FAVORITES, MARKETS } from 'modules/app/constants/views';
+import { TOPIC_PARAM_NAME } from 'modules/app/constants/param-names';
+
+export default class MarketsView extends Component {
```
tinybike: I think I'm repeating myself here, but, `MarketsView` is substantially more complex than before.  Is this solely because state is now being held in the component instead of in Redux, or are there other functional changes buried in here?

src/modules/markets/selectors/markets-totals.js
```diff
@@ -1,30 +1,23 @@
...
   return selectMarketsTotals(store.getState());
 }
 
 export const selectMarketsTotals = createSelector(
   selectAllMarkets,
-  selectFilteredMarkets,
-  selectUnpaginatedMarkets,
-  selectFavoriteMarkets,
-  (allMarkets, filteredMarkets, unpaginatedMarkets, favoriteMarkets) => {
+  (allMarkets) => {
```
tinybike: At a glance, it appears that filtered/unpaginated/favoriteMarkets are no longer memoized; is this correct?  If so, are you concerned about a performance hit?  The markets selectors can get pretty large.

src/modules/transactions/components/transaction-details.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ValueDenomination from 'modules/common/components/value-denomination';
-import Link from 'modules/link/components/link';
+import { Link } from 'react-router-dom';
```
tinybike: Why use `Link` instead of `a` here?

src/modules/trade/helpers/generate-trade.js
```diff
@@ -19,7 +19,9 @@ import store from 'src/store';
  * @param {Object} loginAccount
  * @param {Object} orderBooks Orders for market
  */
-export const generateTrade = memoize((market, outcome, outcomeTradeInProgress, loginAccount, orderBooks) => {
+export const generateTrade = memoize((market, outcome, outcomeTradeInProgress, orderBooks) => {
+  const { loginAccount } = store.getState();
```
tinybike: Is it intentional that `loginAccount` is no longer covered by the memoization?  Generally memoized functions shouldn't have internal calls to `getState`.

tinybike:
Overall I like the react-router structure a lot; it's much less ad hoc than the links system we were using before.

Made a number of efficiency / organization comments above.  Nothing jumped out at me as a show-stopper.  I'm not going to do another full code review of this, so I'll leave it up to you (@stephensprinkle) to address my comments where you it's appropriate / beneficial.  You can go ahead and merge to master when you think it's ready!

package.json
stephensprinkle: Point 1 — yes.  I intentionally made sure our build would not include the entirety of the lib, just the methods we need, that’s why you’ll see import statements like: `import isEqual from ‘lodash/isEqual’;`. It’s got some super useful helper methods that are just faster to import than re-build.  

Point 2 — Good spot, missed that semver caret.

package.json
stephensprinkle: RE: that’s actually some cruft left over from an aborted global state approach to path awareness.

Removing...

package.json
stephensprinkle: No — that’s also cruft from the global state approach.

src/history.js
stephensprinkle: This is related to the inclusion of the history package — also cruft.

src/main.jsx
stephensprinkle: This was no longer needed as a result of the move to react-redux.

src/main.jsx
stephensprinkle: Correct — removing.

src/modules/app/components/routes.jsx
stephensprinkle: Per discussion, I’ll update this function’s name to something that’s a little more clear.

src/modules/app/components/side-bar.jsx
stephensprinkle: 👍

src/modules/app/helpers/parse-query.js
stephensprinkle: Ah, true.

src/modules/auth/components/auth-import.jsx
stephensprinkle: There was an uncaught error coming back from Keythereum if the file or pass was bad I believe, so this catches that and displays an appropriate error.

src/modules/auth/components/auth-view.jsx
stephensprinkle: In this case it’s used for two unique set of user actions.

Creation of an account — `Sign Up`
Authentication — `Login`

stephensprinkle: You’re totally right that we mix the language in the UI though…fixing that up.

src/modules/auth/components/authenticated-route.jsx
stephensprinkle: That’s basically there for convention consistency.  

A little further down you’ll see the use of `Component`, if this wasn’t there, we’d have to do something to the affect of `<props.component…` instead of `<Component…`.

src/modules/common/components/paginator.jsx
```diff
@@ -1,50 +1,242 @@
...
-  isMobile: PropTypes.bool
-};
+import { Link } from 'react-router-dom';
+
+import parseQuery from 'modules/app/helpers/parse-query';
+import makeQuery from 'modules/app/helpers/make-query';
+
+import { PAGINATION_PARAM_NAME } from 'modules/app/constants/param-names';
+
+class Paginator extends Component {
```
stephensprinkle: It’s should effectively be less code now because previously we housed the pagination logic w/in the components that employed the paginator.

I generalized it’s employment and simplified where possible.

stephensprinkle: Parity in functionality btw.

src/modules/market/actions/load-full-market.js
stephensprinkle: I was getting periodic race conditions between when the callback was being called and when the state was actually updated.

src/modules/market/components/market-view.jsx
stephensprinkle: Do you have a preference here?  

I lean towards `marketID`, but I do want to be consistent.

src/modules/markets/components/markets-list.jsx
stephensprinkle: This is mostly due to moving the market info load into this component, such that the market info is only loaded when a market preview is visible.

src/modules/markets/components/markets-view.jsx
stephensprinkle: Yes, functionally this component now manages the conditional loading of markets.

src/modules/markets/selectors/markets-totals.js
stephensprinkle: Per discussion, we haven’t lost memoization (effectively) as it’s now handled by the components and only conditionally updated.

src/modules/transactions/components/transaction-details.jsx
stephensprinkle: Good point, updating.

src/modules/trade/helpers/generate-trade.js
stephensprinkle: 100% right — updating.

PR #925, opened 2017-08-25
Topics header design implementation

smallfx:
Implements "main page" Topics view 🐯
Also includes (very related) fixes to Paginator component.


stephensprinkle:
Looks good overall — a few things to clean up:

- [use `em` for relationally positioned elements + `rem` for globally relationally positioned elements]
- [no margin present on the categories view resulting in cats that overflow]
- [refactor hard coded `px` widths to instead flow with element layout, add padding or margin where necessary]

smallfx:
@stephensprinkle Last commit addresses those concerns with special consideration for `vw` vs `rem`/`em` in certain cases where basing the width on the total screen width was more appropriate for layout fluidity

stephensprinkle:
I ended up rolling back the flex implementation as it ended up not being as clean of an implementation.

PR #928, opened 2017-08-31
New navigation panel + Styleguide / sandbox

smallfx:
Implements `ch2967` 🌟

Also includes a style sandbox page which enables viewing of the new component at `/#/style-sandbox/`

Implemented as a pure stateless component so as to give control to the router versus handling routing / "current view" stuff in local state.
Also accepts an arbitrary number of menu elements and splits menu space evenly between them.


src/modules/app/components/routes.jsx
```diff
@@ -49,6 +49,11 @@ const CreateMarket = asyncComponent({
   loader: () => import(/* webpackChunkName: 'create-market' */ 'modules/create-market/container')
     .then(module => module.default)
 })
+const StyleSandbox = asyncComponent({
+  moduleName: 'StyleSandbox',
+  loader: () => import(/* webpackChunkName: 'topics' */ 'modules/style-sandbox/components/style-sandbox/style-sandbox')
```
stephensprinkle: Update the chunk name to `style-sandbox`.

src/modules/style-sandbox/components/style-sandbox/style-sandbox.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+import NavPanel from 'modules/common/components/nav-panel/nav-panel';
+import AccIcon from 'modules/common/components/nav-account-icon';
+
+const StyleSandbox = (p) => {
+  const navPanelProps = {
+    flipNav: true,
+    items: [
+      { title: 'hello1', iconComponent: AccIcon, onClick: () => alert('one click') },
+      { title: 'hello2', iconComponent: AccIcon, active: true },
```
stephensprinkle: No need for `active` when deriving from `location`

src/modules/style-sandbox/components/style-sandbox/style-sandbox.jsx
```diff
@@ -0,0 +1,32 @@
...
+      { title: 'hello5', iconComponent: AccIcon },
+      { title: 'hello6', iconComponent: AccIcon }
+    ]
+  };
+
+  return (
+    <div className={Styles.StyleSandbox}>
+      <div className={Styles['StyleSandbox__nav-panel-wrap']}>
+        <NavPanel {...navPanelProps}>
+          Placeholder - content will come when appropriate component is available
```
stephensprinkle: Content should be a sibling of nav.  (Per our discussion)

src/modules/common/components/nav-panel/nav-panel.jsx
```diff
@@ -0,0 +1,34 @@
...
+            key={`${item.title}${ind - p.items.length}`}
+            onClick={item.onClick}
+          >
+            <Icon className={Styles.NavPanel__icon} />
+            {item.title}
+          </button>
+        );
+      })}
+    </aside>
+    <div className={Styles.NavPanel__content}>
```
stephensprinkle: Shift this to the parent as a sibling of the nav component. (Per discussion)

src/modules/common/components/nav-panel/nav-panel.jsx
```diff
@@ -0,0 +1,34 @@
...
+const NavPanel = p => (
+  <div className={Styles.NavPanel}>
+    <aside className={Styles.NavPanel__controls}>
+      {p.items && p.items.map((item, ind) => {
+        const Icon = item.iconComponent;
+        return (
+          <button
+            className={classNames(
+              {
+                [Styles.NavPanel__control]: true,
```
stephensprinkle: Instead of setting this here, you could do something to the effect of: 

`classNames(`${Styles.NavPanel__control}`, {…})`

src/modules/common/components/nav-panel/nav-panel.styles.less
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.NavPanel__icon {
+  margin-right: 1.5rem;
+  max-height: 2.25rem;
+  max-width: 2.25rem;
+  vertical-align: middle;
+}
+
+.NavPanel__content {
```
stephensprinkle: No longer needed.

stephensprinkle:
Looks good overall — I’ve added a few points to address via review comments + the inclusion of mobile layout (not originally in scope).

src/modules/common/components/nav-panel/nav-panel.jsx
```diff
@@ -0,0 +1,34 @@
...
+                [Styles['NavPanel__control--active']]: item.active
+              })}
+            key={`${item.title}${ind - p.items.length}`}
+            onClick={item.onClick}
```
stephensprinkle: Instead of passing through an `onClick` property, you can just have a method that updates the query string param for this specific nav (will also need another param) via `history.push`

src/modules/common/components/nav-panel/nav-panel.jsx
```diff
@@ -0,0 +1,34 @@
...
+  <div className={Styles.NavPanel}>
+    <aside className={Styles.NavPanel__controls}>
+      {p.items && p.items.map((item, ind) => {
+        const Icon = item.iconComponent;
+        return (
+          <button
+            className={classNames(
+              {
+                [Styles.NavPanel__control]: true,
+                [Styles['NavPanel__control--active']]: item.active
```
stephensprinkle: Do a comparison against the `location`’s nav param instead of having it passed as a property.

src/modules/common/components/nav-panel/nav-panel.jsx
```diff
@@ -4,46 +4,65 @@ import PropTypes from 'prop-types'
...
-            {item.title}
-          </button>
-        )
+const NavPanel = (p) => {
+  const selectedNavId = decodeURIComponent(parseQuery(p.location.search)[NAVPANEL_ID_PARAM_NAME] || '')
+
+  const toggleNavId = (id) => {
+    const searchParams = makeQuery({
+      ...(parseQuery(p.location.search)),
+      [NAVPANEL_ID_PARAM_NAME]: encodeURIComponent(id)
```
stephensprinkle: No need for `encodeURIComponent` here.  That gets handled via `makeQuery`.

stephensprinkle:
There are still some unaddressed comments.

smallfx:
@stephensprinkle Github still showing a comment that was addressed (by your commit, actually) because I force-pushed after a rebase...
Everything should be good now, though!

smallfx:
@stephensprinkle ok - should be *actually* good to go now!

PR #929, opened 2017-08-31
3030 market preview styles

sharkcrayon:
Finished CH #3030.

Also:

- Removed EmDash component.
- Did some global style adjustments (got rid of font-weight and din-pro mixins)
- Adding an SVG component so we can keep all small SVGs in a single file.


src/modules/market/components/market-properties/market-properties.jsx
```diff
@@ -0,0 +1,73 @@
...
+        </li>
+        <li>
+          <span>Expires</span>
+          <span>{ p.endDate.formatted }</span>
+        </li>
+      </ul>
+      <div>
+        { p.isLogged && p.toggleFavorite &&
+          <button
+            className={classNames(Styles.MarketProperties__favorite, { [`${Styles.favorite}`]: p.isFavorite })}
```
stephensprinkle:
```
[`${Styles.favorite}`]
```
can also just be expressed as `[Styles.favorite]`

stephensprinkle:
Also include the mobile layout w/in the scope of this task.

PR #933, opened 2017-09-03
Add a loading spinner to the topics view

jfarid27:
I would like to add this loading spinner to the topics page to help with the User Experience. When the code is loading, the UI says "No Topics Available", which is pretty confusing to me since there may be topics, but they just haven't loaded yet. A user may see "No Topics Available" while having a slow connection, and assume there really are no topics available. Adding a spinner is better UX since at least the User knows something is happening in the background, and they should wait.

Also, 1ca4616 allows me to pass linter checks on pull. When I run npm lint, this error shows: 
```bash
/Users/jalil/Programs/augur/src/modules/markets/reducers/event-markets-map.js
  10:95  error  Arrow function should not return assignment  no-return-assign
```
This commit is a fix for that, where I just use a reduce statement to aggregate the unique items instead of a stateful filter.

### DIFF
--- a/src/modules/topics/components/topics-view.jsx
+++ b/src/modules/topics/components/topics-view.jsx
@@ -6,6 +6,7 @@ import { Link } from 'react-router-dom';
 import NullStateMessage from 'modules/common/components/null-state-message';
 import TopicRows from 'modules/topics/components/topic-rows';
 import Paginator from 'modules/common/components/paginator';
+import Spinner from 'modules/common/components/spinner';
 import Branch from 'modules/branch/components/branch';
 import FilterSort from 'modules/filter-sort/container';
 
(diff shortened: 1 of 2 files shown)


src/modules/topics/components/topics-view.jsx
```diff
@@ -149,16 +151,19 @@ export default class TopicsView extends Component {
 
   updateFilteredItems(filteredTopics) {
     this.setState({
+      loaded: true,
```
stephensprinkle: The class method `updateFilteredItems` actually gets called almost immediately after mount from within the `FilterSort` component, as a result this spinner most likely would never display.  

A more ideal solution would be to set the `loaded` state in reaction to the actual topics request to augur.

jfarid27:
@stephensprinkle I'll check it out! Thanks.

PR #935, opened 2017-09-05
Sidebar animations + keywords re-implementation

smallfx:
Takes care of ch3493 wherein keyword filtering + selection has changed as a result of the routing refactor. 🎶
Re-implements broken styling + animation, as well as the new route-based keyword selection.　✅


src/modules/app/components/inner-nav/inner-nav.jsx
```diff
@@ -5,6 +5,13 @@ import { mobileMenuStates } from 'modules/app/components/app/app'
 
 import Styles from 'modules/app/components/inner-nav/inner-nav.styles'
 
+import _, { isEqual } from 'lodash';
```
stephensprinkle: Lodash is a fairly large lib, so only import the parts you need.

In this case, you’d want to do `import isEqual from ‘lodash/isEqual’` — this will keep us from having the _entire_ lodash lib included, but rather only `isEqual` and it’s dependencies.

src/modules/app/components/inner-nav/inner-nav.jsx
```diff
@@ -18,15 +25,90 @@ export default class InnerNav extends Component {
...
+    .concat(selectedKeywords)
+    .uniq()
+    .slice(0,50)
+    .map(keyword => ({
+      name: keyword,
+      isSelected: (selectedKeywords || []).indexOf(keyword) !== -1
+    }))
+    .value();
+
+    console.log(filteredKeywords);
```
stephensprinkle: Feel free to remove this.

stephensprinkle:
Also go ahead and fix up the conflicts.

src/modules/app/components/inner-nav/menu-item.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import classNames from 'classnames'
+
+import Styles from 'modules/app/components/inner-nav/inner-nav.styles'
+
+const MenuItem = p => (
+  <li
+    className={classNames({
+      [Styles['InnerNav__menu-item']]: true,
```
stephensprinkle: When you have something that is _always_ true, set it via the first arg of `classNames`.

PR #938, opened 2017-09-15
Fix left-over events and properly dispose P5 when unmounting

smallfx:
Takes care of ch3761 🌟


src/modules/common/components/graph-background/graph-background.jsx
```diff
@@ -32,6 +32,11 @@ export default class GraphBG extends Component {
     return false
   }
 
+  componentWillUnmount() {
+    this.p5Renderer.remove()
```
stephensprinkle: Ah, looks like that’s what I was missing.

PR #939, opened 2017-09-17
#ut account load funding test

bthaile:
Created first unit test, yea!

tests newly updated account/actions/load-funding-history.js


tinybike:
Looks good, thanks! :+1:

PR #942, opened 2017-09-25
Fix background sizing bug

smallfx:
Wanted to fix this w/o relying on polling for element resizes, so it's just CSS minus setting the canvas size to the `<body>` width in JS.
The issue was that there was no screen resize event, though the container effectively resized because of the sidebar animation.


stephensprinkle:
Looks like width changes are now resolved…but height changes look to be causing issues now.

PR #954, opened 2017-10-13
Accounts Withdraw view

smallfx:
Implements `ch3012`, Withdraw view as part of the new Accounts views 🌟


stephensprinkle:
I’m getting an odd behavior when the account sub-nav will slide out when I traverse to the account view un-logged — do you see this as well?

PR #956, opened 2017-10-14
New Account views

smallfx:
Implements `ch3005`, `ch3023` and `ch3012` wherein the new Account views are detailed 🐹


src/modules/new-account/components/account-deposit/account-deposit.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+import Styles from 'modules/new-account/components/account-deposit/account-deposit.styles'
+
+export default class AccountDeposit extends Component {
+  static propTypes = {
+    address: PropTypes.string.isRequired
+  }
+
+  componentDidMount() {
+    const clipboard = new Clipboard('#copy_address') // eslint-disable-line
```
stephensprinkle: Disabling a line is totally fine, but also include the rule you’re disabling.

src/modules/common/components/input-dropdown/input-dropdown.jsx
```diff
@@ -61,7 +61,7 @@ class InputDropdown extends Component {
     return (
       <div
         ref={(InputDropdown) => { this.refInputDropdown = InputDropdown }}
-        className={Styles.InputDropdown}
+        className={classNames(Styles.InputDropdown, (p.className || ''))}
```
stephensprinkle: No need for the conditional here — `classNames` will automatically filter out falsy values.

PR #963, opened 2017-10-19
4214 reporting: closed page

sharkcrayon:


src/modules/reporting/components/reporting-header/reporting-header.jsx
```diff
@@ -28,8 +31,10 @@ export default class ReportingHeader extends Component {
       <article className={Styles.ReportingHeader}>
         <div className={Styles.ReportingHeader__header}>
           <div>
-            <h1 className={Styles.ReportingHeader__heading}>Reporting</h1>
-            <span className={Styles.ReportingHeader__endDate}>Reporting cycle ends { s.endDate }</span>
+            <h1 className={Styles.ReportingHeader__heading}>{p.heading}</h1>
+            { p.showReportingEndDate &&
```
stephensprinkle: Should we also include the cycle graph as conditionally displayed?

Feels odd removing the date, but leaving the cycle graph.  I might be missing something however.

stephensprinkle: On second thought — I’m going to leave this here and we’ll raise it during our discussion w/ Tara.

Per the mocks it’s there (which is where I lean), but I’d be curious to have a discussion before we make a change here.

src/modules/routes/components/routes/routes.jsx
```diff
@@ -27,7 +27,7 @@ const Routes = p => (
     <AuthenticatedRoute path={makePath(VIEWS.TRANSACTIONS)} component={COMPONENTS.Transactions} />
     <AuthenticatedRoute path={makePath(VIEWS.CREATE_MARKET)} component={COMPONENTS.CreateMarket} />
     <AuthenticatedRoute path={makePath(VIEWS.REPORTING)} component={COMPONENTS.Reporting} />
-    <AuthenticatedRoute path={makePath(VIEWS.REPORTING_CLOSED)} component={COMPONENTS.Reporting} />
+    <AuthenticatedRoute path={makePath(VIEWS.REPORTING_CLOSED)} component={COMPONENTS.ReportingClosed} />
```
stephensprinkle: This should be an authenticated sub-view under the reporting view.

src/modules/routes/constants/components.js
```diff
@@ -86,6 +86,12 @@ export const Reporting = asyncComponent({
     .then(module => module.default)
 })
 
+export const ReportingClosed = asyncComponent({
+  moduleName: 'ReportingClosed',
+  loader: () => import(/* webpackChunkName: 'reporting' */ 'modules/reporting/containers/closed')
```
stephensprinkle: Update the webpack chunk name to `reporting-closed`.

sharkcrayon:
This should be ready to go now 👍

PR #964, opened 2017-10-20
3228 v3 Notifications

JohnDanz:
Clubhouse Issue: https://app.clubhouse.io/augur/story/3228/notifications

This ticket adds Notifications back to v3. 

Only one task wasn't met, the clear notifications button. This doesn't appear to be anywhere in the mocks so I've omitted it for the time being.


src/env.json
```diff
@@ -3,7 +3,7 @@
   "gethWebsocketsURL": "ws://127.0.0.1:8546",
   "augurNodeUrl": "ws://127.0.0.1:9001",
   "networkID": null,
-  "autoLogin": false,
+  "autoLogin": true,
```
stephensprinkle: Don’t commit changes made to env.json

src/assets/styles/_colors.less
```diff
@@ -17,6 +17,7 @@
 @color-white: #ffffff;
 @color-almostblack: #1e1a31;
 @color-red: #d0021b;
+@color-transparent: transparent;
```
stephensprinkle: Also remove the var from `_colors_OLD`

src/env.json
JohnDanz: Woops, oversight, I'll make a checklist so i don't forget this next time

src/assets/styles/_colors.less
JohnDanz: done in the next push

src/modules/app/components/side-nav/side-nav.jsx
```diff
@@ -92,11 +98,28 @@ export default class SideNav extends Component {
...
+                  <span className={Styles['item-title']}>{item.title}</span>
                 </Link>
               </li>
             )
           })}
+          {logged && mobile &&
+            <li
+              key="notifications"
+            >
+              <a
```
stephensprinkle: For semantics sake, change this to a `button` since it’s only performing an action + not navigating.

src/modules/notifications/components/notification.jsx
```diff
@@ -1,9 +1,12 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
-import classNames from 'classnames'
+// import classNames from 'classnames'
```
stephensprinkle: Little cleanup here.

PR #967, opened 2017-10-23
3903 reporting form

sharkcrayon:


src/modules/common/less/form.less
```diff
@@ -0,0 +1,277 @@
...
+  margin: auto;
+  max-width: 62.75rem;
+  padding: 3.75rem 0 9.125rem;
+}
+
+.Form__error {
+  background-color: @color-purple;
+  border-radius: 3px;
+  color: @color-white;
+  font-size: 0.625rem;
```
stephensprinkle: I’ve noticed a few locations where font size is being explicitly defined.  All occurrences should be using the shared font-size vars.  While not as rigid, line-height (assuming congruence to font size), should also use the vars for easier refactors to sizing going forward.

src/modules/common/less/confirm-table.less
```diff
@@ -0,0 +1,70 @@
+@import (reference) '~assets/styles/shared';
+
+.Confirm {
+  color: @color-white;
+  width: 36rem;
+}
+
+.Confirm__heading {
+  &:extend(.caps);
+  &:extend(.caps--large);
```
stephensprinkle: Can just use the `.caps—large` mixing here.

src/modules/common/less/confirm-table.less
```diff
@@ -0,0 +1,70 @@
...
+    display: flex;
+    justify-content: space-between;
+
+    &:not(:last-child) {
+      margin-bottom: 1rem;
+    }
+
+    > span:first-child {
+      &:extend(.caps);
+      &:extend(.caps--small);
```
stephensprinkle: Can just use the `.caps—small` mixing here as well.

src/modules/common/less/form.less
```diff
@@ -0,0 +1,277 @@
...
+
+      &.active::after {
+        display: block;
+      }
+    }
+  }
+}
+
+.Form__radio-buttons--per-line {
+  &:extend(.Form__radio-buttons all);
```
stephensprinkle: ooo — I looked up `all`, liking it.

src/modules/reporting/components/reporting-report/reporting-report.jsx
```diff
@@ -0,0 +1,310 @@
...
+    this.prevPage = this.prevPage.bind(this)
+    this.nextPage = this.nextPage.bind(this)
+    this.validateIsMarketValid = this.validateIsMarketValid.bind(this)
+    this.validateOutcome = this.validateOutcome.bind(this)
+    this.validateNumber = this.validateNumber.bind(this)
+    this.validateStake = this.validateStake.bind(this)
+  }
+
+  prevPage() {
+    const newStep = this.state.currentStep <= 0 ? 0 : this.state.currentStep - 1
```
stephensprinkle: You could collapse this into the setState argument as well: 

this.setState({ currentStep: this.state.currentStep <= 0 ? 0 : this.state.currentStep - 1 })

stephensprinkle: ^ This is just a code style comment — not required, just a terser way to express the same thing.

src/modules/reporting/components/reporting-report/reporting-report.jsx
```diff
@@ -0,0 +1,310 @@
...
+        <MarketPreview
+          {...p.market}
+          isLogged={p.isLogged}
+          location={p.location}
+          history={p.history}
+          cardStyle="single-card"
+          buttonText="View"
+        />
+        <article className={FormStyles.Form}>
+          { s.currentStep === 0 &&
```
stephensprinkle: Go ahead and split the steps into separate components — just looking at what’s there, they could simply be pure.

PR #973, opened 2017-11-01
3989 transaction card

sharkcrayon:
This includes code from John's 3985-ptflo-performance-graph branch (since this is branched off of that).


src/modules/portfolio/components/performance-graph/performance-graph.jsx
```diff
@@ -0,0 +1,277 @@
...
+    this.performanceGraph.redraw()
+  }
+
+  render() {
+    // const p = this.props
+    const s = this.state
+
+    return (
+      <section
+        className={Styles.PerformanceGraph}
```
stephensprinkle: These can be on the same line when only one attribute is being set and it’s not too verbose.  

Style critique mainly.

src/modules/portfolio/components/performance-graph/performance-graph.jsx
```diff
@@ -0,0 +1,277 @@
...
+        >
+          <div
+            className={Styles['PerformanceGraph__SortBar-title']}
+          >
+            Profits/losses
+          </div>
+          <div
+            className={Styles['PerformanceGraph__SortBar-dropdowns']}
+          >
+            <Dropdown default={s.graphTypeDefault} options={s.graphTypeOptions} onChange={this.changeDropdown} />
```
stephensprinkle: The convention is to line break when more than one attribute is present.

PR #998, opened 2017-11-09
Account Header Tuning

JohnDanz:
https://app.clubhouse.io/augur/story/4493/graph-should-be-bottom-aligned-account-header

More account header work to try and get it more in line with the mock. the CH issue was for aligning the charts, however I adjusted scaling max down to 5.5 from 7, and adjusted layout to try and get the currency lined up properly with the charts as displayed in the mocks.


src/env.json
```diff
@@ -3,7 +3,7 @@
   "gethWebsocketsURL": "ws://127.0.0.1:8546",
   "augurNodeUrl": "ws://127.0.0.1:9001",
   "networkID": null,
-  "autoLogin": false,
+  "autoLogin": true,
```
stephensprinkle: For shame

PR #1012, opened 2017-11-15
4565 transactions

bthaile:
Backend work only:
backend work to populate transactionData. verified getting data from augur-node for load-account-history. added add-transactions methods to be called from existing methods to flatten transactional data. 
fixed up a few broken unit tests for update-account-trades-data.
Added transaction types to be used on front-end to figure out how to display each transaction.


stephensprinkle:
Per our discussion yesterday, you’re on the right track here.  

I’m going to go ahead and close this PR w/out merging in anticipation of the completed implementation incoming.

PR #1013, opened 2017-11-15
fixed a stylign issue in account header, fixed rep not loading correctly

JohnDanz:
some quick fixes that i noticed while working on other tickets. 

Fixed update-assets to actually correctly make the REP balance call so we actually get our balance displayed ;)

Then fixed the conditional in account header that was skipping the rep view scaling causing it to stay tiny.


stephensprinkle:
There are some broken tests here.

stephensprinkle:
Still failing :/

PR #1016, opened 2017-11-16
4469-categories-congruence

JohnDanz:
https://app.clubhouse.io/augur/story/4469/check-vertical-spacing-of-category-components-for-congruence-to-mocks

Updated the categories desplay to have a bigger hero font, remove the ellipse after "bet on" and increased the length of the hero text underline as it wasn't wide enough. Please review and use the mock data i have in the topics container to test, then remove the mock data ;)

### DIFF
--- a/src/assets/styles/_typography.less
+++ b/src/assets/styles/_typography.less
@@ -34,6 +34,7 @@ h1 {
 @font-size-rather-large: 1.5rem; // 24px
 @font-size-extra-large: 1.875rem; // 30px
 @font-size-massive: 2.43rem; // 39px
+@font-size-super-massive: 4rem; // 64px
 
 //  MIXINS
 .align-to-icon {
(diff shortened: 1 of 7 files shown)


stephensprinkle:
One refinement to the PR — have the pagination remain vertically positioned, even if there aren’t three rows of results.

stephensprinkle:
This approach unfortunately won’t work re: positioning of the pagination as it will result in the pagination rendering ‘beneath’ the categories if the window height is too small.

There are a couple different ways to approach this, but the ultimate effect we need here is for the pagination to remain stationary vertically, no matter the number of results on a page.

One way to absolutely ensure responsive yet stable positioning is to fill the view w/ ‘empty’ categories that are not visible but still take up space for desktop.  On mobile, you can let the paginator component float based on results length…as it will effectively always be on the bottom.

PR #1022, opened 2017-11-18
4565 transactions

bthaile:
Creating PR to get it in to make merging easier before the second coming of christ.
consume data from augur-node, pulling to get user's
- transfers
- orders
- trades
- market creations
Added whitespace in property names for meta properties. fixed wordwrap in meta.

Still need to get all the verbage for thee different transactions and wire in date filter. I think those will be better in different PRs.


src/modules/auth/actions/load-account-history.js
```diff
@@ -1,13 +1,14 @@
...
 import { clearReports } from 'modules/reports/actions/update-reports'
-import { loadAccountTrades } from 'modules/my-positions/actions/load-account-trades'
-import { loadBidsAsksHistory } from 'modules/bids-asks/actions/load-bids-asks-history'
+import { loadUserTradingHistory } from 'modules/my-positions/actions/load-account-trades'
 import { loadCreateMarketHistory } from 'modules/create-market/actions/load-create-market-history'
 import { loadFundingHistory } from 'modules/account/actions/load-funding-history'
 import { loadReportingHistory } from 'modules/my-reports/actions/load-reporting-history'
 import syncUniverse from 'modules/universe/actions/sync-universe'
-import { updateTransactionsOldestLoadedBlock } from 'modules/transactions/actions/update-transactions-oldest-loaded-block'
+/* import { updateTransactionsOldestLoadedBlock } from 'modules/transactions/actions/update-transactions-oldest-loaded-block' */
```
stephensprinkle: Minor thing, but when you comment out one line, just do two prepended `/`s

src/modules/auth/actions/load-account-history.js
```diff
@@ -25,78 +26,56 @@ export const loadAccountHistory = (loadAllHistory, triggerTransactionsExport) =>
     dispatch(syncUniverse())
   }
 
-  if (registerBlock && oldestLoadedBlock && oldestLoadedBlock !== registerBlock) {
-    const options = {}
-    if (!loadAllHistory) {
-      options.toBlock = oldestLoadedBlock === blockchain.currentBlockNumber ? oldestLoadedBlock : oldestLoadedBlock - 1
+  // TODO: figure out where this data comes from and if it is needed
```
stephensprinkle: They’re not explicitly needed any longer — they’re a carry over from the old chunked loading system from on-chain.

src/modules/markets/actions/load-markets-info.js
```diff
@@ -21,3 +21,13 @@ export const loadMarketsInfo = (marketIDs, callback = logError) => (dispatch, ge
     callback(null, marketsData)
   })
 }
+
+export const loadMarketsInfoOnly = (marketIDs, callback = logError) => (dispatch, getState) => {
```
stephensprinkle: Can combine this w/ `loadMarketsInfo`

src/modules/portfolio/containers/transactions.js
```diff
@@ -1,16 +1,24 @@
...
 import TransactionsList from 'modules/portfolio/components/transactions/transactions'
 
 const mapStateToProps = state => ({
-  transactions: dummyTransactions
+  transactions: selectTransactions(state),
+  transactionsLoading: state.transactionsLoading,
+})
+
+const mapDispatchToProps = dispatch => ({
+  // TODO: need to get parameters from configuation?
```
stephensprinkle: How I would do this is have this method consume some args which are passed via a call make w/in the component.

stephensprinkle: Such as on the change of a select element — call this method w/ the selected option.

src/modules/markets/actions/load-markets-info.js
bthaile: I created the only version because the loadMarktsInfo caused a cascade of other api calls that would just collect data not needed for transactions.

src/modules/portfolio/containers/transactions.js
bthaile: How do you want to cascade changes to date filter dropdown?
global event? or is there a cool react way for component to component comm?

bthaile: The date dropdown is in performance-graph.jsx, I'm not sure how transactions.jsx would get it's state. I don't see a parent-child relationship?

PR #1025, opened 2017-11-19
4565 trans clean up

bthaile:
cleaned up unused code. Wired up component sending in a begin date (earliestCreationTime) into augur-node transaction methods. Default to 24Hrs if not set

Added passing in beginDate and endDate. endDate not used/set currently. What is the best way to wire up augur/src/modules/portfolio/components/performance-graph/performance-graph.jsx to tell augur/src/modules/portfolio/components/transactions/transactions.jsx the date-range in the dropdown?


stephensprinkle:
What is the best way to wire up augur/src/modules/portfolio/components/performance-graph/performance-graph.jsx to tell augur/src/modules/portfolio/components/transactions/transactions.jsx the date-range in the dropdown?

Note to self: Addressed via call.

stephensprinkle:
Easy to switch from `_` as opposed to `-` in your branch names on your end?  
If so, move over the dashes instead.

PR #1028, opened 2017-11-25
symbol added to augur-node api for get-account-transfer-history

bthaile:
augur-node api get-account-transfer-history added symbol property. Use this new property to identify what was transferred.


tinybike:
LGTM!

PR #1032, opened 2017-12-04
4698 event listeners

JohnDanz:
https://app.clubhouse.io/augur/story/4698/wire-up-augur-node-listeners
This is going to be an on-going task as there are certain listeners that we can't update yet, like Universe Forked, and there are other listeners that im having trouble testing like the order listeners. 

This PR contains some of the event listeners hooked up and ready/tested.
This PR removes the categorical market description parsing as it's no longer required
This PR fixes Create Market to work with some new params that got changed around


src/env.json
```diff
@@ -3,7 +3,7 @@
   "gethWebsocketsURL": "ws://127.0.0.1:8546",
   "augurNodeUrl": "ws://127.0.0.1:9001",
   "networkID": null,
-  "autoLogin": false,
+  "autoLogin": true,
```
stephensprinkle: The never ending battle continues.

src/modules/app/actions/listen-to-updates.js
```diff
@@ -1,101 +1,118 @@
...
-        // dispatch(syncUniverse())
+        dispatch(syncBlockchain())
+        dispatch(syncUniverse())
       },
       onRemoved: (block) => {
         console.log('block removed:', block)
-        // dispatch(syncBlockchain())
-        // dispatch(updateAssets())
-        // dispatch(syncUniverse())
+        dispatch(syncBlockchain())
```
stephensprinkle: We’ll have to expand this listener as simply updating the block + universe won’t accommodate all changes of a removed block.

PR #1034, opened 2017-12-07
Feature/4555 Load Market Info within Market

pgebheim:


src/modules/market/components/market-view/market-view.jsx
```diff
@@ -1,25 +1,63 @@
...
 export default class MarketView extends Component {
+  static propTypes = {
+    isConnected: PropTypes.bool.isRequired,
+    location: PropTypes.object.isRequired,
+    loadFullMarket: PropTypes.func.isRequired,
+    universe: PropTypes.object.isRequired
+  }
+
+  static loadFullMarket(marketId, props, force) {
+    if (!!force || props.marketsData[marketId] === null) {
```
stephensprinkle: Do a `==` here as really we’re checking whether it’s `undefined`, which wouldn’t get hit w/ a strict comparison.

src/modules/market/components/market-view/market-view.jsx
```diff
@@ -1,25 +1,63 @@
...
+      if (marketId && props.isConnected && getValue(props, 'universe.id')) props.loadFullMarket(marketId)
```
stephensprinkle: `isConnected` denotes we have a universe ID.

src/modules/market/components/market-view/market-view.jsx
```diff
@@ -1,25 +1,63 @@
...
+
 export default class MarketView extends Component {
+  static propTypes = {
+    isConnected: PropTypes.bool.isRequired,
+    location: PropTypes.object.isRequired,
+    loadFullMarket: PropTypes.func.isRequired,
+    universe: PropTypes.object.isRequired
+  }
+
+  static loadFullMarket(marketId, props, force) {
```
stephensprinkle: It’s probably better to explicitly consume an options object here and explicitly define the properties from the props object.

src/modules/market/containers/market-view.js
```diff
@@ -18,27 +20,18 @@
...
-// })
-//
-// // const Market = withRouter(connect(null, null)(MarketView))
-//
-// export default Market
+const mapStateToProps = state => ({
+  isLogged: state.isLogged,
+  isConnected: state.connection.isConnected,
+  universe: state.universe,
+  marketsData: state.marketsData
```
stephensprinkle: Since we don’t actually use `marketsData` in this view, I would personally use mergeProps, derive the marketID from the location, then pass down something like `isMarketLoaded`, which has it’s value as `state.marketsData[marketID] != null`.

src/modules/bids-asks/actions/load-one-outcome-bids-or-asks.js
```diff
@@ -11,11 +11,11 @@ const loadOneOutcomeBidsOrAsks = (marketID, outcome, orderTypeLabel, callback =
...
   dispatch(updateIsFirstOrderBookChunkLoaded(marketID, outcome, orderTypeLabel, false))
-  augur.trading.getOpenOrders({ marketID, outcome, orderType: orderTypeLabel }, (err, openOrders) => {
+  augur.trading.getOrders({ marketID, outcome, orderType: orderTypeLabel }, (err, orders) => {
     if (err) return callback(err)
-    if (openOrders != null) {
-      // TODO verify that openOrders is the correct shape for insertion
-      dispatch(insertOrderBookChunkToOrderBook(marketID, outcome, orderTypeLabel, openOrders))
+    if (orders != null) {
+      // TODO verify that orders is the correct shape for insertion
+      dispatch(insertOrderBookChunkToOrderBook(marketID, outcome, orderTypeLabel, orders))
```
stephensprinkle: Ideally we load _all_ orders, instead of by side, but I get that this is how it’s currently implemented.

src/modules/market/components/market-view/market-view.jsx
pgebheim: Oh duh! Good catch

src/modules/bids-asks/actions/load-one-outcome-bids-or-asks.js
pgebheim: Should we file an issue to fix?

stephensprinkle: Yeah, we should…but it’s probably post MVP.  I’ll toss it into Seadragon — Unscoped + let Serena make the call.

src/modules/market/containers/market-view.js
pgebheim: This now happens in mergeProps

src/modules/market/components/market-view/market-view.jsx
```diff
@@ -20,6 +33,21 @@ export default class MarketView extends Component {
...
+  componentWillMount() {
+    if (this.props.isConnected && !this.props.isMarketLoaded) {
+      this.props.loadFullMarket()
+    }
+  }
+
+  componentWillUpdate(nextProps, nextState) {
+    if (
+      (this.props.isConnected === false && nextProps.isConnected === true) ||
+      this.props.marketId !== nextProps.marketId
```
stephensprinkle: We don’t actually need this conditional here as this won’t change w/out re-routing/re-mounting.

src/modules/market/components/market-view/market-view.jsx
```diff
@@ -9,6 +10,18 @@ import MarketTrading from 'modules/market/containers/market-trading'
 import Styles from 'modules/market/components/market-view/market-view.styles'
 
 export default class MarketView extends Component {
+  static propTypes = {
+    marketId: PropTypes.string.isRequired,
+    isConnected: PropTypes.bool.isRequired,
+    isMarketLoaded: PropTypes.bool.isRequired,
+    location: PropTypes.object.isRequired,
```
stephensprinkle: Don’t need this prop validation or the universe validation.

stephensprinkle:
Just a little light cleanup, but otherwise looks good!

stephensprinkle:
So, there is a little more work to do re: integrating the data into the selector, but I will actually take that on in course w/ my market view wrap up.

PR #1045, opened 2017-12-15
[Cleanup] Rename filenames containing "topic" to "category"

kennsippell:
https://app.clubhouse.io/augur/story/5204/update-all-references-of-topics-categories-keywords-tags


kennsippell:
1. Not exactly sure the procedure to verify this isn't breaking anything. I've started manual testing, but pretty broad-reaching change and I'm pretty scared.
1. Git seems to be struggling to `move` files instead of `add/delete`. I restarted the change and made a separate commit each time I renamed a folder, but it doesn't seem to have solved the problem. Let me know if you care to preserve that, and I'll try some more things.
1. Potentially an ugly merge for anybody with outstanding changes. Not sure on best time to do this. Not sure if I should do the "Keywords>Tags" change together in one PR or if it best to keep them separate.

kennsippell:
I'm going to break this into smaller changes.  This one just moves files around, but doesn't touch the content.

.... refactoring like this is a slog in javascript

stephensprinkle:
That’s a totally sane approach.

Re: tooling — I know IntelliJ’s IDEs have that built in, so rename a var in one place + it’s updated _everywhere_.

PR #1046, opened 2017-12-17
[Cleanup] Rename 'keyword' to 'tag'

kennsippell:
https://app.clubhouse.io/augur/story/5204/update-all-references-of-topics-categories-keywords-tags


kennsippell:
Much more straightforward in comparison to https://github.com/AugurProject/augur/pull/1045. Most of the uses of `keyword` appear to be in unused code @ modules/filter-sort. I believe I did a basic sanity check of all other components involved in this change.

The only remaining instance of 'keyword' in the codebase after this is the filter-sort constant. I believe this is currently unused and probably best unchanged. `export const FILTER_SEARCH_PARAM = 'keywords'`

PR #1050, opened 2017-12-18
5068 null state styling

bthaile:
updated styling for null markets, update mobile styling, added link to markets when no positions.


src/modules/my-positions/actions/load-account-positions.js
```diff
@@ -1,14 +1,28 @@
...
 export const loadAccountPositions = (options, callback = logError) => (dispatch, getState) => {
   const { universe, loginAccount } = getState()
   if (!loginAccount.address) return callback(null)
   augur.trading.getUserTradingPositions({ ...options, account: loginAccount.address, universe: universe.id }, (err, shareBalances) => {
     if (err) return callback(err)
     if (shareBalances == null) return callback(null)
-    dispatch(updateAccountPositionsData(shareBalances, options.marketID))
-    callback(null, shareBalances)
+    // TODO: need better way to consolidate by marketID
+    const marketIDs = shareBalances.reduce((collapsed, position) => {
```
stephensprinkle: The convention we’ve adopted in the CB for reduce args is `(p, <item>, …)`.

src/modules/my-positions/actions/load-account-positions.js
```diff
@@ -1,14 +1,28 @@
...
+      if (!collapsed[position.marketID]) {
+        collapsed[position.marketID] = []
+      }
+      collapsed[position.marketID].push(position)
+      return collapsed
+    }, {})
+    dispatch(loadMarketsInfo(Object.keys(marketIDs).slice(), () => {
```
stephensprinkle: Shouldn’t be a need for a `slice` here.

src/modules/portfolio/components/positions/positions.jsx
```diff
@@ -1,58 +1,85 @@
...
-      orderCancellation={p.orderCancellation}
-      location={p.location}
-      history={p.history}
-      positionsDefault={false}
-    />
-  </section>
-)
+export default class Positions extends Component {
+  static propTypes = {
+    location: PropTypes.object.isRequired,
```
stephensprinkle: Make sure to put any optional props last for easier readability.

src/modules/portfolio/components/positions/positions.jsx
```diff
@@ -1,58 +1,85 @@
...
+    history: PropTypes.object.isRequired,
+    openPositionMarkets: PropTypes.array.isRequired,
+    reportingMarkets: PropTypes.array.isRequired,
+    closedMarkets: PropTypes.array.isRequired,
+    closePositionStatus: PropTypes.object.isRequired,
+    scalarShareDenomination: PropTypes.object.isRequired,
+    orderCancellation: PropTypes.object.isRequired,
+    registerBlockNumber: PropTypes.number,
```
stephensprinkle: Make sure to put any optional props last for easier readability.

src/modules/my-positions/actions/load-account-positions.js
bthaile: Is this what you mean? is `position` considered a <singular item name> ?
    
const marketIDs = positions.reduce((p, position) => {
      if (!p[position.marketID]) {
        p[position.marketID] = []
      }
      p[position.marketID].push(position)
      return p
    }, {})

stephensprinkle: Yup, that’s what I was driving towards.

I’m going to do a slight cleanup here to simplify this method, give me one sec.

stephensprinkle: K, pushed!

src/modules/portfolio/components/positions/positions.jsx
```diff
@@ -1,58 +1,85 @@
...
 
-export default Positions
+  render() {
+    const p = this.props
+    return (
+      <section>
+        <Helmet>
+          <title>Positions</title>
+        </Helmet>
+        { p.marketsCount !== 0 &&
```
stephensprinkle: You can make this just one conditional that wraps those three components.  You’ll also need to wrap them in a div, so you might need to also update the styling a bit.

src/modules/portfolio/components/positions/positions.jsx
```diff
@@ -1,58 +1,85 @@
...
+          closePositionStatus={p.closePositionStatus}
+          scalarShareDenomination={p.scalarShareDenomination}
+          orderCancellation={p.orderCancellation}
+          location={p.location}
+          history={p.history}
+          positionsDefault={false}
+        /> }
+        { p.marketsCount === 0 &&
+          <div className={PortfolioStyles.NoMarkets__container} >
+            <span>You don&apos;t have any positions.</span>
```
stephensprinkle: You’re safe to just use `’` here.

PR #1053, opened 2017-12-21
4149 positions items

bthaile:
augur-node integration for positions. pull in user positions and get user orders. inject user orders into order book. Then populate market outcomes with userOpenOrders, had to shape data in user-open-orders.

Had an issue with augur/src/modules/markets/selectors/markets-all.js `selectOrderBooksState` didn't get orderBooks from state. its acting like it never knew data was added to state.orderBooks, I modified user-open-orders to get orderBooks from state which pulled as orderBooks was updated.


src/modules/bids-asks/actions/load-account-orders.js
```diff
@@ -1,18 +1,28 @@
...
+  augur.trading.getOrders({ ...options, creator: loginAccount.address, universe: universe.id }, (err, orders) => {
     if (err) return callback(err)
     if (orders == null || Object.keys(orders).length === 0) return callback(null)
     const marketIDs = Object.keys(orders)
     // TODO: consolidate all the getting of maket infos for load account history
-    dispatch(loadMarketsInfoOnly(marketIDs.slice(), () => {
+    dispatch(loadMarketsInfo(marketIDs.slice(), () => {
       dispatch(addOpenOrderTransactions(orders))
-      // TODO update user's open orders
+      async.forEachOfSeries(marketIDs, (marketID, id, next) => {
```
stephensprinkle: No need that I see for the async lib here, refactor this to just use the `forEach` method.  

While I’m not opposed to using external libs when a real utility is needed, when a base javascript comparable suffices, the preference is towards that in the CB.

src/modules/bids-asks/actions/load-account-orders.js
```diff
@@ -1,18 +1,28 @@
...
 export const loadAccountOrders = (options, callback = logError) => (dispatch, getState) => {
   const { universe, loginAccount } = getState()
-  augur.trading.getOrders({ ...options, account: loginAccount.address, universe: universe.id }, (err, orders) => {
+  augur.trading.getOrders({ ...options, creator: loginAccount.address, universe: universe.id }, (err, orders) => {
     if (err) return callback(err)
     if (orders == null || Object.keys(orders).length === 0) return callback(null)
     const marketIDs = Object.keys(orders)
     // TODO: consolidate all the getting of maket infos for load account history
-    dispatch(loadMarketsInfoOnly(marketIDs.slice(), () => {
+    dispatch(loadMarketsInfo(marketIDs.slice(), () => {
```
stephensprinkle: No need to slice here — you can just pass in the `marketIDs`.

src/modules/market/selectors/market.js
```diff
@@ -36,10 +36,10 @@ import { submitReport } from 'modules/reports/actions/submit-report'
 import store from 'src/store'
 
 import selectAccountPositions from 'modules/user-open-orders/selectors/positions-plus-asks'
-import selectUserOpenOrders from 'modules/user-open-orders/selectors/user-open-orders'
+import { selectUserOpenOrders } from 'modules/user-open-orders/selectors/user-open-orders'
 import selectUserOpenOrdersSummary from 'modules/user-open-orders/selectors/user-open-orders-summary'
 
-import { selectPriceTimeSeries } from 'modules/market/selectors/price-time-series'
+// import { selectPriceTimeSeries } from 'modules/market/selectors/price-time-series'
```
stephensprinkle: 👍

src/modules/market/selectors/market.js
```diff
@@ -36,10 +36,10 @@ import { submitReport } from 'modules/reports/actions/submit-report'
 import store from 'src/store'
 
 import selectAccountPositions from 'modules/user-open-orders/selectors/positions-plus-asks'
-import selectUserOpenOrders from 'modules/user-open-orders/selectors/user-open-orders'
+import { selectUserOpenOrders } from 'modules/user-open-orders/selectors/user-open-orders'
```
stephensprinkle: I’m not 100% settled on the location of this selector, it’s fine for now.

src/modules/user-open-orders/selectors/user-open-orders.js
```diff
@@ -11,17 +11,32 @@ import { CLOSE_DIALOG_CLOSING } from 'modules/market/constants/close-dialog-stat
 
 import { formatNone, formatEtherTokens, formatShares } from 'utils/format-number'
 
+export function selectAllUserOpenOrderMarkets() {
+  const { loginAccount, orderBooks } = store.getState()
```
stephensprinkle: This is an older pattern that actually is still present in a few locations, but shouldn’t be implemented going forward.

Take a look at how we’ve structured selectors with the `createSelector` convention and refactor for congruence to that structure.  

Also, break out this method into a separate file.

src/modules/user-open-orders/selectors/user-open-orders.js
```diff
@@ -11,17 +11,32 @@ import { CLOSE_DIALOG_CLOSING } from 'modules/market/constants/close-dialog-stat
...
 /**
- *
+ * Pulls off existing order book in state
  * @param {String} outcomeId
- * @param {{buy: object, sell: object}} marketOrderBook
+ * @param {String} marketID
  *
  * @return {Array}
  */
-export default function (outcomeId, marketOrderBook) {
```
stephensprinkle: I’m not a fan of this change — let’s go ahead and roll this all back.  

We should be trying to diagnose the core issue of why it wasn’t picking up the state change prior to refactoring other methods.

src/modules/user-open-orders/selectors/user-open-orders.js
bthaile: fixed to use createSelector pattern.

src/modules/user-open-orders/selectors/user-open-orders.js
bthaile: reverted user-open-orders to orig. 
modified markets-all to use market.js selectMarket instead of assembleMarket

stephensprinkle:
Node 6 was giving that odd yarn issue again — merging on down.

PR #1056, opened 2017-12-22
5252 qrcode resizing

bthaile:
Updated to set the QR Code size based on isMobile or IsMobileSmall. This will remove QR code jumping around when user turns their device. Also removes code.


stephensprinkle:
I’m going to be making some changes to this PR — while this works, we more ideally would like the QR code to be responsive to the element reflows instead of having a fixed size at various breakpoints.

PR #1057, opened 2017-12-26
fixed calls to the api.Universe contract that have changed

JohnDanz:
Fixed some calls that have changed due to contract updates. Previous methods being used have been renamed to fit the new reporting system (reporting window - dispute round, etc)


stephensprinkle:
This can be closed w/out merging, correct?

tinybike:
No, I think these are still needed.

PR #1059, opened 2017-12-30
Env updates and category fixes

tinybike:


tinybike:
Going ahead and merging this to un-break the UI.

PR #1070, opened 2018-01-11
4883 dev deployment docker

bthaile:
created ipfs enabled docker file that exposes two endpoints for testing for dev. Will have dev.augur.net point to 8001 (nginx on top of build dir) and port 80 points to ipfs hosted augur build dir.


package.json
```diff
@@ -12,6 +12,7 @@
     "url": "https://github.com/AugurProject/augur"
   },
   "scripts": {
+    "clean": "rm -rf build; rm -rf node_modules",
```
stephensprinkle: AFAIK, this will fail when run in a win environment.

Ideally use something like `rimraf` for compatibility.

bthaile: updated to use rimraf, great catch.

stephensprinkle:
👍

PR #1076, opened 2018-01-14
Docker ipfs config

bthaile:
changed webpack to create references to relative path instead of absolute path.
added run local environment variables for users to run augur ui without ipfs daemon running.
fixed issue with publishing, needed to $(...) the command
added clean up in docker container to remove ipfs install files


support/ipfs-crontab
```diff
@@ -1 +1 @@
-* */12 * * * /usr/local/bin/ipfs name publish $(cat /augur/ipfs-deploy/NEW_BUILD_HASH) > /var/log/cron.log
+* */24 * * * /usr/local/bin/ipfs name publish $(cat /augur/ipfs-deploy/NEW_BUILD_HASH) > /var/log/cron.log
```
stephensprinkle: Curious why the move to 24 over 12?

support/ipfs-run.sh
```diff
@@ -1,13 +1,17 @@
 #!/bin/bash
 
-ipfs daemon &
-sleep 5s
+if [ "$RUN_LOCAL_ONLY" != "true" ]; then
+  ipfs daemon &
+  sleep 10s
```
stephensprinkle: Minor improvement — you could listen for the status of the ipfs process + pause the script.

webpack.config.js
```diff
@@ -37,7 +37,7 @@ let config = {
     filename: '[name].[chunkhash].js',
     chunkFilename: '[name].[chunkhash].js',
     path: PATHS.BUILD,
-    publicPath: '/'
+    publicPath: ''
```
stephensprinkle: 👍

support/ipfs-crontab
bthaile: I read that using a non default key the default time-to-live is 12 hours, but saw on the API the default is 24 hours, so I changed. Also the publish can throw up a Warning that gets spit out as a ERROR and I wanted to minimize frequency of the message.

support/ipfs-run.sh
bthaile: I went to easy route and put in a sleep. I agree could have put in a do loop to process based on standard out.

webpack.config.js
bthaile: yea, that was fun to find. learned a bit about webpack, cool stuff.

stephensprinkle: Nice!

support/ipfs-crontab
stephensprinkle: Yeah, defaults to 24, but It’s actually configurable as well.  

I’d originally set it to 12 hours to just make sure there was never a gap.

PR #1079, opened 2018-01-16
fixed unit test, resolved stub needed a return value.

bthaile:
changes to ledger unit test to get it passing locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: node_js
 branches:
   only:
     - master
-    - v3
     - seadragon
 
 env:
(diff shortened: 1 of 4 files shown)


stephensprinkle:
These changes don’t cover the changes to the function, but that’s OK in this case as it’s not yet finalized.  

Assuming everything passes I’ll go ahead and merge this down to get things passing and I’ll come back after I’m finished up and get this to full coverage.

bthaile:
I'm not sure why the test passes locally but not in CI. I don't have any uncommitted files.

stephensprinkle:
Updating these tests now for coverage + successful execution in 7/6

PR #1082, opened 2018-01-18
moved default port 80 to serve up build directory, now have port 8001…

bthaile:
swapped ports in dockerfile. Exposing port 8080 didn't work out so I have 8001 as pass through to 8080. Default port 80 now serves up the build directory.

Added running docker container with all ports mapped. Added some general docker management commands.


stephensprinkle:
Looks good!  Made some small copy refinements for readability.

PR #1084, opened 2018-01-18
5484-liquidity-data-bug

JohnDanz:
fixed a bug that was causing the market outcome liquidity info to not display on the market cards.

https://app.clubhouse.io/augur/story/5484/markets-page-debug-why-liquidity-data-isn-t-display-on-market-cards


stephensprinkle:
Good catch.

PR #1087, opened 2018-01-22
5547 notification updates

JohnDanz:
https://app.clubhouse.io/augur/story/5547/notifications-style-and-functionality-fixes

Don't actually merge this until @sharkcrayon has weighed in on the position of the mobile null state message. Otherwise this should be good to go. One thing i added that wasn't explicitly in the ticket was a resize/position of the mobile close notification X that appears in the top left corner. I made it match it's position and size to the close X that appears when the side nav is open in mobile. this makes it feel more consistent and doesn't feel as jarring.

### DIFF
--- a/src/assets/styles/_box-shadow.less
+++ b/src/assets/styles/_box-shadow.less
@@ -1,3 +1,3 @@
 // TODO -- there are a few occurances of box shadow
 // 	For consistency, need to build up a mixin
-@box-shadow-notifications: 0 2px 4px 0 rgba(0,0,0,0.50);
+@box-shadow-notifications: 0 2px 4px 0 rgba(0, 0, 0, 0.5);
--- a/src/assets/styles/_box-shadow.less
+++ b/src/assets/styles/_box-shadow.less
@@ -1,2 +1,3 @@
 // TODO -- there are a few occurances of box shadow
 // 	For consistency, need to build up a mixin
+@box-shadow-notifications: 0 2px 4px 0 rgba(0,0,0,0.50);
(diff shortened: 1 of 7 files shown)


src/assets/styles/_box-shadow.less
```diff
@@ -1,2 +1,3 @@
 // TODO -- there are a few occurances of box shadow
 // 	For consistency, need to build up a mixin
+@box-shadow-notifications: 0 2px 4px 0 rgba(0, 0, 0, 0.5);
```
stephensprinkle: Noice

src/modules/app/components/app/app.jsx
```diff
@@ -355,7 +369,7 @@ export default class AppView extends Component {
       // ensure origami fold-out moves perfectly with submenu
       origamiScalar = Math.max(0, (subMenu.scalar + mainMenu.scalar) - 1)
     }
-
+    // console.log('rendering app', p)
```
stephensprinkle: 😢

src/modules/notifications/components/notifications-view.styles.less
```diff
@@ -2,18 +2,18 @@
...
   padding: 0.25em;
   position: absolute;
-  right: 0.25rem;
+  right: 1.8rem;
   text-align: left;
   top: 3rem;
-  width: 20em;
-  z-index: 1000;
+  width: 22.75rem;
+  z-index: @mask-modal;
```
stephensprinkle: Can you confirm that this won’t cause the modal to reside beneath the notifications when present?

@mask-modal is intended for full view modals only.

JohnDanz: switched this to mask-above. it's still mask-modal on mobile however as it is intended to act as a full screen display. Is this going to need it's own z-index instead so that we can show a overall page modal?

PR #1089, opened 2018-01-23
5660 linux build

tinybike:
Updated babel-eslint and webpack-dev-middleware versions as suggested by https://github.com/salomonelli/best-resume-ever/issues/188.


stephensprinkle:
There’s actually a newer version of `webpack-dev-middleware`, but no changelog and if 2.0.3 is working for you, we’ll stick w/ that.

PR #1092, opened 2018-01-26
5497 mobile style

bthaile:
https://app.clubhouse.io/augur/story/5497/mobile-nav-style-fixes

added space at top of mobile menu, added account balances and remove style to get svg dynamic.


src/modules/app/components/side-nav/side-nav.styles.less
```diff
@@ -72,8 +72,11 @@
 
 @media @breakpoint-mobile {
   .SideNav {
+    display: flex;
+    flex-flow: column nowrap;
     height: 100vh;
     left: -12.5rem;
+    overflow: scroll;
```
stephensprinkle: Update this to `auto`

Auto will only show the scrollbar when it’s needed, whereas scroll will always show it.

stephensprinkle: Sorry, updated the comment prob after you saw it.  

Also update the property to `overflow-y`

src/modules/app/components/side-nav/side-nav.jsx
```diff
@@ -122,6 +123,26 @@ export default class SideNav extends Component {
             </li>
           }
         </ul>
+        {isLogged && isMobile &&
+        <div className={Styles.SideName__space} />
```
stephensprinkle: No need for this — you can apply the flex directly to the `.SideNav__nav`, which should result in the same outcome.

src/modules/app/components/side-nav/side-nav.styles.less
```diff
@@ -84,9 +87,48 @@
...
+  .SideName__space {
+    flex-grow: 3;
+  }
+
+  .SideName__placement {
+    padding: 20% 0 20% 20%;
+  }
+
+  .SideNav__amt {
+    bottom: 0;
```
stephensprinkle: No need for this.

src/modules/app/components/side-nav/side-nav.styles.less
```diff
@@ -84,9 +87,48 @@
...
 
+  .SideNav__nav__separator {
+    background: @color-lightpurple;
+    height: 0.0625rem;
+    margin: 1rem 0 0.5rem;
+    width: 12.125rem;
+  }
+
+  .SideNav__stat-label {
+    color: rgba(255, 255, 255, 0.25);
```
stephensprinkle: Use less’ `fadeout` function w/ a var from our `colors.less` stylesheet.  

Totally OK to set an opacity percentage manually here.

src/modules/app/components/side-nav/side-nav.styles.less
```diff
@@ -104,17 +106,14 @@
...
-    flex-grow: 3;
-  }
-
   .SideName__placement {
     padding: 20% 0 20% 20%;
   }
 
   .SideNav__amt {
-    bottom: 0;
+    align-self: flex-end;
```
stephensprinkle: Ah, wouldn’t have done it that way, but that works!

PR #1093, opened 2018-01-26
h2 was causing issues with gap between bet on and separator, needed t…

bthaile:
h2 was causing issues with gap between bet on and separator, needed to hid if no hero category is available.


stephensprinkle:
Also take a look into + resolve why there is never a category displaying there.

PR #1097, opened 2018-01-30
for some reason binary markets only got 1 outcome data added to state…

bthaile:
for some reason binary markets only got 1 outcome data added to state, so positions and open orders would not show up in portfolio positions


stephensprinkle:
Since we’re only allowing traders to trade against ‘Yes’ in the UI, we probably don’t need this.  

I’m leaving open temporarily per discussion in `dev`, just in case something else surfaces.

stephensprinkle:
Closing as this isn’t necessary in the UI.

PR #1098, opened 2018-01-30
5465 adjust categories

JohnDanz:
https://app.clubhouse.io/augur/story/5465/homepage-categories-should-be-horizontally-centered-when-there-are-less-than-three-categories

Re-adds the Hero Category.

adjusted categories to display as noted in the CH issue. I suggest we eventually revisit this section and use a grid layout or something to improve the quality of the code, but for now MVP is more important ;)


stephensprinkle:
You could simplify this a bit to just conditionally set `p.boundedLength` to either be the pagination max length OR the categories count — that way the same styles can apply as well as have the component be a bit simpler.

PR #1100, opened 2018-02-01
5776 close position

bthaile:
revamped portfolio positions. fix ability to cancel orders and close positions. Still need to fix formatting between the interaction of Positions and Orders in portfolio page. need more testing of different markets and orderbook.

fixed up mobile layout for newly vamped Portfolio Positions page.


stephensprinkle:
A few failing tests — it looks like they might be due to some changes coming down the pipes in the follow-on task, so if so, feel free to comment out temporarily.

PR #1101, opened 2018-02-01
Expose tickSize in market create form

justinbarry:
ref #4828


stephensprinkle:
One minor change — make sure the step size validates against being less than 0.

sharkcrayon:
A couple minor design updates:
- We need to preserve the spacing between the fields. See the "Multiple Choice" market type and the spacing between the Potential Outcomes fields.
- "Range Denomination" can be shortened to "Denomination".

@stephensprinkle Are traders going to understand what ".0001" is? We don't have a label or placeholder text for that field currently, so there's no indication of what it is.

stephensprinkle:
Traders do understand the concept of ‘tick size’, but as per the current implementation of the input field, I wouldn’t expect it to be immediately obvious as to what exactly that field means.  We could set the placeholder to say ‘Tick Size’, which some traders would understand, but ultimately some copy or label w/ tooltip would be definitely be helpful.

stephensprinkle:
@sharkcrayon — just making sure you’re notified re: comment above.

justinbarry:
@sharkcrayon I added a label to tick size but could probably use further refinement.

sharkcrayon:
When I view the Create Market:Outcome page and click on Numerical Range, I'm getting a JS error that says `TypeError: Cannot read property 'length' of undefined`.

Anyone else getting that?

justinbarry:
Just fixed. Unsure why it worked for me yesterday but it def works now.

justinbarry:
Also note the error message runs over the label I added for tick size. How should I address this?
screen shot 2018-02-06 at 9 13 18 am

stephensprinkle:
Per the changes coming in https://app.clubhouse.io/augur/story/5669/update-create-market-error-placement, that should get resolved.

@sharkcrayon — you OK merging this as is under the assumption that 5669 will resolve that layout issue?

sharkcrayon:
@stephensprinkle Yep, good to go!

PR #1110, opened 2018-02-05
5820 market disclaimer

nuevoalex:


src/modules/create-market/components/create-market-form-define/create-market-form-define.jsx
```diff
@@ -97,9 +97,7 @@ export default class CreateMarketDefine extends Component {
         <li className={Styles.CreateMarketDefine__question}>
           <label htmlFor="cm__input--desc">
             <span>Market Question</span>
-            { p.newMarket.validations[p.newMarket.currentStep].description.length &&
-              <span className={StylesForm.CreateMarketForm__error}>{InputErrorIcon}{ p.newMarket.validations[p.newMarket.currentStep].description }</span>
-            }
+            { p.newMarket.validations[p.newMarket.currentStep].description.length }
```
stephensprinkle: That was only used for the logical operator, so you’re safe to remove this as well.

nuevoalex: Ah ok, Will do.

PR #1122, opened 2018-02-09
5881 total rep stake

bthaile:
small PR to use augur-node's convenient method, rest of PR on other CH task

ended up being all there is needed for 5838 as well.
https://app.clubhouse.io/augur/story/5838/dispute-view-header


src/modules/reporting/actions/load-reporting-window-bounds.js
```diff
@@ -3,33 +3,19 @@ import { updateReportingWindowStats } from 'modules/reporting/actions/update-rep
...
-    })
-
-    augur.api.FeeWindow.getEndTime({ tx: { to: reportingWindow } }, (err, endTime) => {
+  augur.augurNode.submitRequest(
+    'getFeeWindowCurrent',
+    {
+      universe: universe.id,
+      reporter: loginAccount.address
+    }, (err, result) => {
       if (err) return
```
tinybike: Hm, should there be a callback here (or at least a `console.error`), so if there is an error it propagates to the caller?  Same question for other `if (err) return` lines.

bthaile: Yes we need to at least log it. If error we wint update fee window properties. I’ll fix

bthaile: added callback to log error

src/env-dev.json
```diff
@@ -5,7 +5,7 @@
     "ws": "ws://127.0.0.1:8546"
   },
   "network-id": 4,
-  "auto-login": true,
+  "auto-login": false,
```
tinybike: Was this changed on purpose?

bthaile: nope not meant to be changed.

bthaile:
removed the change to env-dev.json file.

PR #1125, opened 2018-02-10
Setup Redux dev tools in app

justinbarry:
These changes are required to use this: 
https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd

Signed-off-by: Justin Barry <328965+justinbarry@users.noreply.github.com>


package.json
```diff
@@ -66,6 +66,7 @@
     "react-router-dom": "4.2.2",
     "react-transition-group": "2.2.1",
     "redux": "3.7.2",
+    "redux-devtools-extension": "^2.13.2",
```
stephensprinkle: We lock versions, so exclude the caret.

PR #1126, opened 2018-02-11
Escape hatch. Significant portions left TODO as they are blocked.

nuevoalex:
Many parts are blocked still. Currently only the market owner escape hatch is hooked up correctly. The share escape hatch is more or less implemented but there are some parts waiting on implementation at other layers.

Not sure what the standard process is for PRs where things are fully implemented. Wanted to get feedback and potentially get in in and send the issue to blocked if possible for now.


tinybike:
Heads-up @nuevoalex, there are several test failures in the Travis build.

nuevoalex:
Ah, will fix.

tinybike:
@nuevoalex what are the things you need implemented in other layers to finish this up?  Is it data from augur-node, new contract code, new augur.js bindings, or...?  Generally I find it's most efficient if you can work by issue rather than by layer, so if you can implement whatever is needed yourself in the other layers, I encourage you to do that!

stephensprinkle:
@tinybike @nuevoalex — status on this?

Want to make sure there isn’t anything outstanding for me on this PR.

tinybike:
@nuevoalex needed some changes to augur-node for this to work, which @epheph said he was handling.  Have these augur-node changes happened yet?  If not, suggest that @nuevoalex just makes the changes himself, in general it's not good for PRs to just sit open like this.

PR #1127, opened 2018-02-11
4556-verify-positions

bthaile:
a few issues, race condition: markets load before user logging in with meta mask. (not an issue using unlocked user).  Portfolio positions shows markets without Positions or Orders.  

First saw issue when there are orders on outcome 0 on binary markets. Code not needed, not sure why my-positions relied on user orders to get market ids, when positions.js container merges in user open orders markets.

### DIFF
--- a/src/env-dev.json
+++ b/src/env-dev.json
@@ -5,7 +5,7 @@
     "ws": "ws://127.0.0.1:8546"
   },
   "network-id": 4,
-  "auto-login": true,
+  "auto-login": false,
   "universe": null,
   "debug": {
     "connect": true,
--- a/src/modules/auth/actions/set-login-account.js
+++ b/src/modules/auth/actions/set-login-account.js
@@ -1,4 +1,5 @@
 import { useUnlockedAccount } from 'modules/auth/actions/use-unlocked-account'
+import { loadAccountTrades } from 'modules/my-positions/actions/load-account-trades'
 import logError from 'utils/log-error'
 
 // If there is an available unlocked account, set as the user's sending address.
@@ -6,5 +7,6 @@ export const setLoginAccount = (autoLogin, account, callback = logError) => (dis
   // If autoLogin=true, use an unlocked local Ethereum node (if present)
   if (autoLogin) {
     dispatch(useUnlockedAccount(account, callback))
+    dispatch(loadAccountTrades())
   }
 }
--- a/src/modules/auth/actions/set-login-account.js
+++ b/src/modules/auth/actions/set-login-account.js
@@ -1,5 +1,4 @@
 import { useUnlockedAccount } from 'modules/auth/actions/use-unlocked-account'
-import { loadAccountHistory } from 'modules/auth/actions/load-account-history'
 import logError from 'utils/log-error'
 
 // If there is an available unlocked account, set as the user's sending address.
@@ -7,6 +6,5 @@ export const setLoginAccount = (autoLogin, account, callback = logError) => (dis
   // If autoLogin=true, use an unlocked local Ethereum node (if present)
   if (autoLogin) {
     dispatch(useUnlockedAccount(account, callback))
-    dispatch(loadAccountHistory())
   }
 }
--- a/src/modules/auth/actions/set-login-account.js
+++ b/src/modules/auth/actions/set-login-account.js
@@ -1,5 +1,5 @@
 import { useUnlockedAccount } from 'modules/auth/actions/use-unlocked-account'
-import { loadAccountTrades } from 'modules/my-positions/actions/load-account-trades'
+import { loadAccountHistory } from 'modules/auth/actions/load-account-history'
 import logError from 'utils/log-error'
 
 // If there is an available unlocked account, set as the user's sending address.
@@ -7,6 +7,6 @@ export const setLoginAccount = (autoLogin, account, callback = logError) => (dis
   // If autoLogin=true, use an unlocked local Ethereum node (if present)
   if (autoLogin) {
     dispatch(useUnlockedAccount(account, callback))
-    dispatch(loadAccountTrades())
+    dispatch(loadAccountHistory())
   }
 }
--- a/src/modules/auth/actions/use-unlocked-account.js
+++ b/src/modules/auth/actions/use-unlocked-account.js
@@ -1,6 +1,7 @@
 import { augur } from 'services/augurjs'
 import { loadAccountData } from 'modules/auth/actions/load-account-data'
 import { updateIsLogged } from 'modules/auth/actions/update-is-logged'
+import { loadAccountHistory } from 'modules/auth/actions/load-account-history'
 import isMetaMask from 'modules/auth/helpers/is-meta-mask'
 import logError from 'utils/log-error'
 
@@ -29,6 +30,7 @@ export const useUnlockedAccount = (unlockedAddress, callback = logError) => (dis
       return callback(null)
     }
     dispatch(updateIsLoggedAndLoadAccountData(unlockedAddress))
+    dispatch(loadAccountHistory())
     callback(null)
   })
 }
--- a/src/modules/auth/actions/use-unlocked-account.js
+++ b/src/modules/auth/actions/use-unlocked-account.js
@@ -1,7 +1,7 @@
 import { augur } from 'services/augurjs'
 import { loadAccountData } from 'modules/auth/actions/load-account-data'
 import { updateIsLogged } from 'modules/auth/actions/update-is-logged'
-import { loadAccountHistory } from 'modules/auth/actions/load-account-history'
+import { loadAccountOrders } from 'modules/bids-asks/actions/load-account-orders'
 import isMetaMask from 'modules/auth/helpers/is-meta-mask'
 import logError from 'utils/log-error'
 
@@ -18,6 +18,7 @@ const updateIsLoggedAndLoadAccountData = unlockedAddress => (dispatch) => {
     },
     isUnlocked: true
   }, true))
+  dispatch(loadAccountOrders())
 }
 
 // Use unlocked local address (if actually unlocked)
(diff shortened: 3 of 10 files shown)


src/modules/my-positions/selectors/my-positions.js
```diff
@@ -11,6 +10,5 @@ export default function () {
 export const selectPositionsMarkets = createSelector(
   selectMarkets,
   selectAccountPositionsState,
-  selectAllUserOpenOrderMarkets,
-  (markets, positions, marketIDs) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID) || (marketIDs || {}).find(marketID => market.id === marketID))
+  (markets, positions) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID))
```
tinybike: What is the intended functionality of this selector?  I think it's not working the way you expect it to, since the filter function's return value isn't Boolean.  If you just want to pull out an array of market objects in which the user has positions, I think all you need is `(markets || []).filter(market => (positions || {})[market.id] != null)`, although honestly I'd break this apart into 2 lines for readability and efficiency's sake:
```javascript
(markets, positions) => {
  if (!markets || !positions) return []
  return markets.filter(market => positions[market.id] != null)
}
```

test/auth/actions/set-login-account-test.js
```diff
@@ -21,11 +21,14 @@ describe(`modules/auth/actions/set-login-account.js`, () => {
...
       }
       const UseUnlockedAccount = { useUnlockedAccount: () => {} }
+      const LoadAccountTrades = { loadAccountTrades: () => {} }
       const action = proxyquire('../../../src/modules/auth/actions/set-login-account.js', {
         '../../../services/augurjs': AugurJS,
+        '../../my-positions/actions/load-account-trades': LoadAccountTrades,
         './use-unlocked-account': UseUnlockedAccount
       })
       sinon.stub(UseUnlockedAccount, 'useUnlockedAccount').callsFake(account => ({ type: 'USE_UNLOCKED_ACCOUNT', account }))
+      sinon.stub(LoadAccountTrades, 'loadAccountTrades').callsFake(account => ({ type: 'LOAD_USER_TRADES', data: {} }))
```
tinybike: Suggest using `'LOAD_ACCOUNT_TRADES'` as the `type` for consistency.

src/modules/auth/actions/set-login-account.js
```diff
@@ -1,10 +1,12 @@
 import { useUnlockedAccount } from 'modules/auth/actions/use-unlocked-account'
+import { loadAccountTrades } from 'modules/my-positions/actions/load-account-trades'
 import logError from 'utils/log-error'
 
 // If there is an available unlocked account, set as the user's sending address.
 export const setLoginAccount = (autoLogin, account, callback = logError) => (dispatch, getState) => {
   // If autoLogin=true, use an unlocked local Ethereum node (if present)
   if (autoLogin) {
```
tinybike: Indirectly relevant to this PR, but, it appears the only callsite for `setLoginAccount` is app/actions/init-augur.js:
```javascript
dispatch(setLoginAccount(true, account))
```
Unless I'm misunderstanding something, the first argument should be determined by the user-specified `env.autoLogin` setting, rather than being hardcoded to `true`.

src/modules/auth/actions/set-login-account.js
```diff
@@ -1,10 +1,12 @@
...
     dispatch(useUnlockedAccount(account, callback))
+    dispatch(loadAccountTrades())
```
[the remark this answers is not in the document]
bthaile: I think we are getting away from unlocked account, if I understand how it works, users would need a geth/parity client running locally. I think we should not offer unlocked account as an option for simplicity sake.

Primarily I was concerned with loading user open orders when user isn't logged in, Transactions View calls loadAccountHistory. 

I can make changes if you think we still are going to support unlocked accounts.

src/modules/my-positions/selectors/my-positions.js
bthaile: I agree, I got confirmation form @stephensprinkle that having positions and orders together is left over when data was being pulled from contracts. 

I'll simplify it, can delete this selector and do it single line in caller.

src/modules/auth/actions/set-login-account.js
bthaile: Not comfortable changing it. Not sure the difference between meta mask and other login methods. If I change to use auto-login from env then it doesn't log in even though meta mask is logged in. 

I'll leave to @stephensprinkle and make a clean up task in CH.

https://app.clubhouse.io/augur/story/6582/cleanup-initaugur-takes-true-for-auto-login

src/env-dev.json
```diff
@@ -5,7 +5,7 @@
     "ws": "ws://127.0.0.1:8546"
   },
   "network-id": 4,
-  "auto-login": true,
+  "auto-login": false,
```
tinybike: Is this setting actually doing anything right now?

src/modules/my-positions/selectors/my-positions.js
```diff
@@ -10,5 +10,9 @@ export default function () {
 export const selectPositionsMarkets = createSelector(
   selectMarkets,
   selectAccountPositionsState,
-  (markets, positions) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID))
+  (markets, positions) => {
+    if (!markets || markets.length === 0) return []
+    if (!positions || positions.length === 0) return []
+    return (markets || []).filter(market => positions[market.id] != null)
```
tinybike: I think this can just be:
```javascript
return markets.filter(market => positions[market.id] != null)
```
since you're checking for `!markets` already.

src/modules/my-positions/selectors/my-positions.js
```diff
@@ -10,5 +10,9 @@ export default function () {
 export const selectPositionsMarkets = createSelector(
   selectMarkets,
   selectAccountPositionsState,
-  (markets, positions) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID))
+  (markets, positions) => {
+    if (!markets || markets.length === 0) return []
+    if (!positions || positions.length === 0) return []
```
tinybike: Is `positions` an array or an object?  The previous code had `Object.keys(positions || {})`, which I'm assuming means that it is an object.

src/modules/auth/actions/set-login-account.js
```diff
@@ -1,12 +1,12 @@
...
+import { loadAccountHistory } from 'modules/auth/actions/load-account-history'
 import logError from 'utils/log-error'
 
 // If there is an available unlocked account, set as the user's sending address.
 export const setLoginAccount = (autoLogin, account, callback = logError) => (dispatch, getState) => {
   // If autoLogin=true, use an unlocked local Ethereum node (if present)
   if (autoLogin) {
     dispatch(useUnlockedAccount(account, callback))
-    dispatch(loadAccountTrades())
+    dispatch(loadAccountHistory())
```
tinybike: I think it would make more sense for `loadAccountHistory` to be called at the end of `useUnlockedAccount`.  If the account isn't actually unlocked, we don't want to load the history.

tinybike:
@bthaile tests still failing on Travis.

src/modules/my-positions/selectors/my-positions.js
```diff
@@ -11,6 +10,11 @@ export default function () {
 export const selectPositionsMarkets = createSelector(
   selectMarkets,
   selectAccountPositionsState,
-  selectAllUserOpenOrderMarkets,
-  (markets, positions, marketIDs) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID) || (marketIDs || {}).find(marketID => market.id === marketID))
+  (markets, positions) => {
+    if (!markets || !positions) {
+      return []
+    }
+    (markets || []).filter(market => Object.keys(positions || {})
```
bthaile: Need return here, accidently pushed this

src/env-dev.json
bthaile: I didn't track down how it's used. I've set it to false so I can use metamask and not local unlocked node. When searching the codebase I didn't find any usages.
Apparently it's superstitious behavior. Still recommend setting to false, we need not to accidently use it. 😃

src/modules/my-positions/selectors/my-positions.js
bthaile: positions is keyed array of 0,1 (buy|sell) of arrays of positions with market  id.

src/modules/auth/actions/use-unlocked-account.js
```diff
@@ -29,6 +30,7 @@ export const useUnlockedAccount = (unlockedAddress, callback = logError) => (dis
       return callback(null)
     }
     dispatch(updateIsLoggedAndLoadAccountData(unlockedAddress))
+    dispatch(loadAccountHistory())
```
tinybike: The account history should be loaded whether the user is using MetaMask or a local Ethereum node.  Suggest putting `dispatch(loadAccountHistory())` on line 20, right after the `loadAccountData` dispatch.

src/modules/my-positions/selectors/my-positions.js
```diff
@@ -11,6 +10,11 @@ export default function () {
 export const selectPositionsMarkets = createSelector(
   selectMarkets,
   selectAccountPositionsState,
-  selectAllUserOpenOrderMarkets,
-  (markets, positions, marketIDs) => (markets || []).filter(market => Object.keys(positions || {}).find(positionMarketID => market.id === positionMarketID) || (marketIDs || {}).find(marketID => market.id === marketID))
+  (markets, positions) => {
+    if (!markets || !positions || Object.keys(positions).length === 0) {
+      return []
+    }
+    return (markets || []).filter(market => Object.keys(positions || {})
```
tinybike: This is wrong again.

stephensprinkle:
@tinybike — is this ready for my review?

PR #1132, opened 2018-02-13
6614-market-create-date

bthaile:
timestamp needs to be in seconds unix timestamp. This looks like format-date thought it was in  milliseconds format. The Create market UI was showing year 1970 and causing market creates to fail. At least that is 1 reason the fail.

https://app.clubhouse.io/augur/story/6614/market-create-date-control-not-setting-date-in-state-correctly
currently in unscheduled, I thought this was a priority so made the code change based on input from @epheph and @justinbarry 

pushed as much date stuff as I could into format-date. 2 instances of using moment() which needed * 1000 to render the correct date, event though timestamp should have been in seconds.

### DIFF
--- a/src/modules/auth/actions/transfer-funds.js
+++ b/src/modules/auth/actions/transfer-funds.js
@@ -31,7 +31,7 @@ export function transferFunds(amount, currency, toAddress) {
               id: `onSent-${tx.hash}`,
               title: `Transfer Ether -- Pending`,
               description: `${amount} ETH -> ${trimString(to)}`,
-              timestamp: parseInt(getCurrentDateTimestamp(), 10),
+              timestamp: getCurrentDateTimestamp(),
             }))
           },
           onSuccess: (tx) => {
@@ -39,7 +39,7 @@ export function transferFunds(amount, currency, toAddress) {
               id: `onSent-${tx.hash}`,
               title: `Transfer Ether -- Success`,
               description: `${amount} ETH -> ${trimString(to)}`,
-              timestamp: parseInt(getCurrentDateTimestamp(), 10),
+              timestamp: getCurrentDateTimestamp(),
             }))
             dispatch(updateAssets)
           },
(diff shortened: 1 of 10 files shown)


src/utils/format-date.js
```diff
@@ -35,7 +35,7 @@ export function formatDate(d) {
     formattedLocal: `${months[date.getMonth()]} ${date.getDate()}, ${date.getFullYear()} ${localTimeTwelve.join(':')} ${localAMPM} (UTC ${localOffset})`, // local time
     formattedLocalShort: `${months[date.getMonth()]} ${date.getDate()}, ${date.getFullYear()} (UTC ${localOffset})`, // local time
     full: date.toUTCString(),
-    timestamp: date.getTime() / 1000
+    timestamp: date.getTime()
```
stephensprinkle: `date.getTime()` returns millis, so it looks like this was what you wanted prior to this change as it will now set `timestamp` to the full milliseconds from epoc.

stephensprinkle: My best guess as to what is happening here is the date passed into this helper method already has the millis trimmed.

bthaile: Yes, I agree

There are so many / 1000 in the UI that I got a bit confused.

src/utils/format-date.js
```diff
@@ -51,7 +51,7 @@ function getTwelveHour(time) {
   return time
 }
 
-export function convertUnix(integer) {
+export function convertUnix(integer = 0) {
```
stephensprinkle: Be a little more verbose on this function name, just a suggestion: `convertUnixToFormatedDate`

stephensprinkle:
Also, got some failing tests.

src/modules/auth/actions/transfer-funds.js
```diff
@@ -30,15 +31,15 @@ export function transferFunds(amount, currency, toAddress) {
               id: `onSent-${tx.hash}`,
               title: `Transfer Ether -- Pending`,
               description: `${amount} ETH -> ${trimString(to)}`,
-              timestamp: parseInt(Date.now() / 1000, 10),
+              timestamp: parseInt(getCurrentDateTimestamp(), 10),
```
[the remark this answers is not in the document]
bthaile: Ok, I’ll make that change

src/utils/format-date.js
bthaile: I’ll make that change and fix unit test, :facepalm:

PR #1134, opened 2018-02-14
Run unit tests inside pre-push hook

justinbarry:


justinbarry:
Do we want these to be silent? It seems like **_some_** feed back would be awesome.

stephensprinkle:
We’d set to silent so the terminal wasn’t spammed, but it would still throw if any issue arose.

Merging as I’m OK w/ this change.  I personally still do `—no-verify` on my feature branches.

PR #1135, opened 2018-02-14
6197 bug bounty filter

nuevoalex:
Configuration and support for only requesting markets created by the bug bounty market uploader.


stephensprinkle:
I’ll take this one on as I have the context.

PR #1138, opened 2018-02-15
5782-norm-trans-market

bthaile:
fixed issue where bad transaction formatted blank create market entries gets added to users transaction from convert log to transaction when market is created. 
fixed unit test. 
saw that augur-node sends log.market we were expecting log.marketId, so changed and added comment


src/modules/transactions/reducers/transactions-data.js
```diff
@@ -24,6 +25,12 @@ export default function (transactionsData = DEFAULT_STATE, action) {
         return p
       }, {})
     case CLEAR_TRANSACTION_DATA:
+      return Object.keys(transactionsData).reduce((p, transactionID) => {
+        if (transactionsData[transactionID].status === PENDING) {
+          p[transactionID] = transactionsData[transactionID]
+        }
+        return p
+      }, {})
     case RESET_STATE:
```
bthaile: clear transaction data is called when browser is redirected to transactions or user goes to transactions, need to retain transactions in pending state.

PR #1140, opened 2018-02-16
Create Tooltip Component

sharkcrayon:
https://app.clubhouse.io/augur/story/6459/create-tooltip-component

I added an example tooltip. You can view it on the Reporting: Dispute view by hovering over "Example Text" (which will be changed to "Participate" in a future story).

One difference from the mock: this center of this tooltip aligns with its corresponding text, rather than being bottom aligned. I don't think this is something we want to fool with since we're using a plugin for positioning, so I'm happy with letting that visual constraint go.

Note: I adjusted some of the structure and styles within `reporting-header.jsx` when adding "Example Text". The content here (the Dispute Window progress bar, etc) hasn't been updated to the newest mocks. As a result, my updates may cause a merge conflict if work on those updates has already started. (The related story for bringing that content in line with the new mocks is "Ready for Deploy", but I could not find the matching PR in either Open or Closed PRs. https://app.clubhouse.io/augur/story/5838/dispute-view-header)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "react-modal-dialog": "4.0.7",
     "react-redux": "5.0.6",
     "react-router-dom": "4.2.2",
+    "react-tooltip": "^3.4.0",
     "react-transition-group": "2.2.1",
     "redux": "3.7.2",
     "redux-devtools-extension": "2.13.2",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "react-modal-dialog": "4.0.7",
     "react-redux": "5.0.6",
     "react-router-dom": "4.2.2",
+    "react-tooltip": "^3.4.0",
```
stephensprinkle: Make sure to exclude the caret when adding packages.

PR #1141, opened 2018-02-16
Removing all code related to the Account:Export view.

sharkcrayon:
A couple of important notes:

I wasn't able to thoroughly test the result of removing this code. 
- While I can log into the site locally with MetaMask and there are no errors showing there as a result of removing Export-related code, I'm not sure that that's relevant because the Export view wasn't present in the Account subnav when logging in with Metamask to begin with (previous to these changes).
- The Export view _is_ present when logging in with Airbitz on the *dev* site, which makes me want to test these changes when logged in with Airbitz locally to make sure no errors are thrown. However, I'm not currently able to do that (should I be able to with our current dev environment?).

Also, I'm questioning my original assumption (that I doubled check with Jack) that we don't need the Export view because we no longer have Keystore. Is that correct? Should it be possible to export the private key when logged in through Airbitz? If so, this PR/branch should be thrown out.


sharkcrayon:
I see also that I failed to update some tests. I'll update those if you confirm that we want to move forward with this.

stephensprinkle:
RE: 
MM — We don’t get access to the private key, so that’s correct that there is no export displayed.
Airbitz — Currently, the app only runs on the `Rinkeby` network, so you’ll need to connect to the Rinkeby hosted nodes to test that.  Airbitz login _should_ still be working.

We’re still going to allow users to login w/ a local hosted node, but IMO we still shouldn’t expose the ability to export in the UI (They already have the ability to export via their ETH client).

sharkcrayon:
I agree. Ok, moving forward with ripping out the Export view.

Tests are updated. This should be good to go!

stephensprinkle:
You’ve got a merge conflict here as well — I’ll fix up real quick then merge down.

PR #1143, opened 2018-02-16
WIP: Run tests placed in the src directory

justinbarry:


justinbarry:
This is a WIP. The glob I added isn't picking up suffix-less tests in the test folder.

PR #1144, opened 2018-02-17
6223-simple-report-routing

bthaile:
added component routing for report on market.

https://app.clubhouse.io/augur/story/6223/reporting-form-add-component-and-routing

### DIFF
--- a/src/modules/routes/components/routes/routes.jsx
+++ b/src/modules/routes/components/routes/routes.jsx
@@ -20,6 +20,7 @@ const Routes = p => (
     <Route path={makePath(VIEWS.MARKET)} component={COMPONENTS.Market} />
     <Route path={makePath(VIEWS.CONNECT)} component={COMPONENTS.Connect} />
     <Route path={makePath(VIEWS.CREATE)} component={COMPONENTS.Create} />
+    <Route path={makePath(VIEWS.REPORT)} component={COMPONENTS.Report} />
     <AuthenticatedRoute path={makePath(VIEWS.FAVORITES)} component={COMPONENTS.Markets} />
     <AuthenticatedRoute path={makePath(VIEWS.MY_POSITIONS)} component={COMPONENTS.Portfolio} />
     <AuthenticatedRoute path={makePath(VIEWS.MY_MARKETS)} component={COMPONENTS.Portfolio} />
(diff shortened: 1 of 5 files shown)


src/modules/routes/components/routes/routes.jsx
```diff
@@ -20,6 +20,7 @@ const Routes = p => (
     <Route path={makePath(VIEWS.MARKET)} component={COMPONENTS.Market} />
     <Route path={makePath(VIEWS.CONNECT)} component={COMPONENTS.Connect} />
     <Route path={makePath(VIEWS.CREATE)} component={COMPONENTS.Create} />
+    <Route path={makePath(VIEWS.REPORT)} component={COMPONENTS.Report} />
```
sharkcrayon: I'm thinking this should be an AuthenticatedRoute. And for that matter, unrelated to this PR, the route for `CREATE` (one line above) should be too. @stephensprinkle is that right?

src/modules/routes/constants/views.js
```diff
@@ -10,6 +10,7 @@ export const CONNECT = 'connect'
 export const CREATE = 'create'
 export const CATEGORIES = 'categories'
 export const REPORTING = 'reporting'
+export const REPORT = 'report'
```
sharkcrayon: You've been renaming `REPORTING_REPORT` just to `REPORT` generally, right? So here, instead of adding this line, I think you actually want to edit line 46 (`export const REPORTING_REPORT = 'reporting-report'`).

src/modules/routes/constants/components.js
```diff
@@ -79,3 +79,9 @@ export const Reporting = asyncComponent({
   loader: () => import(/* webpackChunkName: 'reporting' */ 'modules/reporting/components/reporting-view/reporting-view')
     .then(module => module.default)
 })
+
+export const Report = asyncComponent({
```
sharkcrayon: If I remember correctly, this file is meant for just top-level views. If that's correct, you shouldn't be adding this here (the code directly above, for `moduleName: 'Reporting',` should be sufficient). @stephensprinkle Is that correct?

src/modules/routes/components/routes/routes.jsx
stephensprinkle: For `REPORT`, yes, it should be authenticated.  

`CREATE` is actually for account creation…which ultimately is going away, but for now it should be a regular route.

src/modules/routes/constants/components.js
stephensprinkle: That’s correct + been the convention thus far.  What these do is establish Webpack breakpoints, so this is where you get discrete files in the build.  

Let’s keep to this convention for now, but later we can potentially establish additional breakpoints.  For now, just be declarative when it comes to routes within their respective components.

PR #1148, opened 2018-02-19
4785 order integration

JohnDanz:
https://app.clubhouse.io/augur/story/4785/integration-validation-market-order-form

Also includes this CH issue: https://app.clubhouse.io/augur/story/4779/add-deposit-funds-button-to-market-trading-form-on-mobile

When testing, you will need to have approved the Cash contract (my next ticket is adding UI for this). to check your allowance run this in the console:
```
augur.api.Cash.allowance({ _owner: state.loginAccount.address, _spender: augur.contracts.addresses[augur.rpc.getNetworkID()].Augur });
```
if that doesn't return a massive number, it's most likely 0, so run this in the console to approve:
```
augur.accounts.approveAugur(state.loginAccount.address, state.loginAccount.auth, function(err, res) { console.log(err, res); });
```

further, the canned markets for scalar have bunk orderbooks that are outside of the tickSize/lowest share size. I would suggest either modifying them to be more realistic in the `augur.js/scripts/dp/data/canned-markets.js` file before you deploying them locally on your own docker node or creating a scalar market to test against so that it's accurately setup.


stephensprinkle:
🎉 — will target to give this a full look over tomorrow morning.

PR #1160, opened 2018-02-22
Fixing account balance update issue

adrake33:
This is intended to fix https://app.clubhouse.io/augur/story/5746/update-account-balances-on-tx-in, where the user's account balance (ETH, REP, etc.) in the UI is not getting updated when certain transactions occur. The problem seems to stem from some if statements in the code like:

```
if (log.sender === getState().loginAccount.address) {
  // Update user's account balance
}
```

`log.sender` is not getting set, which is causing these checks to fail.  I have confirmed that my fix works for market creation, but I am currently unable to use the UI to create/cancel/fill orders or submit reports, so I have not tested the other if conditions yet.

@epheph Can you please confirm whether these changes look correct based on what Augur Node is emitting for these events?


src/modules/app/actions/listen-to-updates.js
```diff
@@ -44,7 +44,7 @@ export function listenToUpdates(history) {
           console.log('MarketCreated:', log)
           // augur-node emitting log.market from raw contract logs.
           dispatch(loadMarketsInfo([log.market]))
-          if (log.sender === getState().loginAccount.address) {
```
epheph: Does this code listen for both log additions and removals?

adrake33: Hmm, good question. If a log is removed, will Augur Node emit that using `MarketCreated`? How should we be checking for that?

epheph: Well, we've been pretty inconsistent about this, because when we insert data, we build up large objects to represent the insert. When we remove, we often only need transactionHash/logIndex so we dont build up the same object to emit on removal.

adrake33: Ok, so it sounds like `log.marketCreator` won't exist in such cases, correct? Do you think this is something that should be changed on the augur-node side?

epheph: Probably, i imagine we'll need to go out of our way to fetch certain data during a log removal, but we haven't known what it is (and i don't imagine it'll be the entire identical object, just an object with the key details to trigger a refresh)

epheph: So actually, market creator WILL exist in this case, it comes directly off the log, and we emit the entire log

https://github.com/AugurProject/augur-node/blob/6c5dbcf45093a123dd0002f08f6f7f9aeb3c6741/src/blockchain/log-processors/market-created.ts#L44

adrake33: Oh, alright. But what about the other ones below (log.filler, log.orderCreator, & log.reporter)?

epheph: here's a good example of a problem: https://github.com/AugurProject/augur-node/blob/6c5dbcf45093a123dd0002f08f6f7f9aeb3c6741/src/blockchain/log-processors/order-filled/index.ts#L99

orderCreator comes off the row for order-filled, and we emit the row we're going to insert during log addition
```
        augurEmitter.emit("OrderFilled", tradeData);
```

but only the log (without orderCreator) on removal:
```
  augurEmitter.emit("OrderFilled", log);
```

I could fix, but is that the only thing you need added?

adrake33: As far as I'm aware, just the references to `log.sender` need to be fixed. (I also just noticed that I missed one for `TradingProceedsClaimed`.) However, I haven't gone through each of the log types in this file to confirm, so there may be other ones that need to be updated as well.

adrake33: Is the log addition for`TradingProceedsClaimed`not finished yet on augur-node?  I see the following in `trading-proceeds-claimed.ts`: 

```
export function processTradingProceedsClaimedLog(db: Knex, augur: Augur, log: FormattedEventLog, callback: ErrorCallback): void {
  console.log("TODO: TradingProceedsClaimed");
  console.log(log);
  callback(null);
}
```

epheph: yes, unimplemented. want it?

src/modules/app/actions/listen-to-updates.js
```diff
@@ -97,7 +97,7 @@ export function listenToUpdates(history) {
           dispatch(updateOutcomePrice(log.marketId, log.outcome, new BigNumber(log.price, 10)))
           dispatch(updateMarketCategoryPopularity(log.market, log.amount))
           const { address } = getState().loginAccount
-          if (log.sender === address || log.owner === address) {
```
epheph: This value gets emitted as 'creator' not 'orderCreator' (it actually differs from DB in this way, translated in the object)

https://github.com/AugurProject/augur-node/blob/6c5dbcf45093a123dd0002f08f6f7f9aeb3c6741/src/blockchain/log-processors/order-filled/index.ts#L54

src/modules/app/actions/listen-to-updates.js
adrake33: If it's not much trouble, sure. Ideally, I'd like to get all of these log types resolved with this PR.

src/modules/app/actions/listen-to-updates.js
adrake33: Good catch! I'll update this.

src/modules/app/actions/listen-to-updates.js
epheph: OK, writing and testing the getter sometimes takes a bit, not sure how long you wanted to wait before merging. I could probably start on it tomorrow.

adrake33: That's ok, I can hold off on merging this until it's ready.

PR #1161, opened 2018-02-22
4353 escape hatch view

nuevoalex:
When the escape hatch is pulled the UI will be blocked by a modal. Any assets the user has available which can be drained in this case are loaded, and presuming any exist they may send txs which will send those funds out to them.

Kind of a large PR so don't hesitate to ask questions.


justinbarry:
To test run the following in the console as the blessed user:

`augur.api.Controller.emergencyStop({
    onSent: (err, success) => {}, 
    onSuccess: (err, success) => {}, 
    onFailure: (err, success) => {}
})`

`augur.api.Controller.release({
    onSent: (err, success) => {}, 
    onSuccess: (err, success) => {}, 
    onFailure: (err, success) => {}
})`

PR #1162, opened 2018-02-22
6244 fix error style

bthaile:
fixed error styling based on feedback in [ch6244]

### DIFF
--- a/src/modules/reporting/components/reporting-report-form/reporting-report-form.styles.less
+++ b/src/modules/reporting/components/reporting-report-form/reporting-report-form.styles.less
@@ -6,24 +6,6 @@ ul.ReportingReportForm__fields {
   justify-content: flex-start;
 }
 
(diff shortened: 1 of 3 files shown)


src/modules/reporting/components/reporting-report-form/reporting-report-form.styles.less
```diff
@@ -6,6 +6,24 @@ ul.ReportingReportForm__fields {
   justify-content: flex-start;
 }
 
+.ReportingReport__Form__err {
+  color: @color-negative;
```
justinbarry: Can we simplify here by extending FormStyles.Form__error?

justinbarry: The BEM model as I understand it is Block__Element--modification. Is error a modification or an element here?

PR #1167, opened 2018-02-23
Account section mobile styles.

sharkcrayon:
https://app.clubhouse.io/augur/story/6546/account-view-mobile-style-updates

### DIFF
--- a/src/modules/account/components/account-deposit/account-deposit.styles.less
+++ b/src/modules/account/components/account-deposit/account-deposit.styles.less
@@ -11,7 +11,7 @@
   > h1 {
     color: @color-darkpurple;
     display: inline-block;
-    font-size: 1.875rem;
+    font-size: @font-size-extra-large;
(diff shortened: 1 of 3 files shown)


sharkcrayon:
@justinbarry Have you gotten a chance to take a look at this?

justinbarry:
@sharkcrayon Are child selectors (of elements) considered an anti-pattern in BEM?
https://developer.mozilla.org/en-US/docs/Web/CSS/Child_selectors

sharkcrayon:
@justinbarry No. If you try to put a class on _everything_, it quickly becomes silly. I try to stick to AirBnB's rule that if you have a selector that's nested more than three deep (`.Example__content > p > a`) then you're doing it wrong. https://github.com/airbnb/css#nested-selectors

And, _almost almost almost_ always, if you're nesting, you should be using child selectors.

PR #1170, opened 2018-02-23
Input aren't necessarily defined when unpacking params

justinbarry:
Just found this guy while testing pr #1161


epheph:
@justinbarry and i were talking about this and discovered that this is the first transaction called by the UI that does not have any input data

tinybike:
Hm, can `inputs` actually be undefined?  I'm looking over our ABI, seems like functions w/ no parameters still have `inputs` defined as `[]`.  (It might be a good idea to have a default assignment anyway though, in case of malformed transactions.)

PR #1172, opened 2018-02-25
5913 account approval

JohnDanz:
https://app.clubhouse.io/augur/story/5913/account-approval


tinybike:
Hm, it seems like it would be preferable to have this dialog either on the account page, or as a prompt when you attempt to make your first trade / do anything that requires an approval.  IMO it shouldn't be a blocking modal on login; users should be able to browse the site without sending an approval tx.

src/modules/auth/actions/approve-account.js
```diff
@@ -0,0 +1,22 @@
...
+      if (err) callback(err)
+      callback(null, allowance)
+    })
+  }
+}
+
+export function approveAugur(callback = logError) {
+  return (dispatch, getState) => {
+    const { loginAccount } = getState()
+    augur.accounts.approveAugur(loginAccount.address, loginAccount.auth, callback)
```
tinybike: Two comments:
1. `approveAugur` should use augur.js's normal API format, which is uniform between different functions, and includes all public-facing functions.
2. `loginAccount.auth` should be `loginAccount.meta`

src/modules/auth/actions/load-account-data.js
```diff
@@ -11,4 +17,12 @@ export const loadAccountData = account => (dispatch) => {
   dispatch(updateLoginAccount(account))
   dispatch(loadAccountPositions())
   dispatch(updateAssets())
+  dispatch(checkAccountAllowance((err, allowance) => {
+    if (err) return console.log(err)
```
tinybike: Suggest using the usual callback-with-default-`logError` instead of hardcoding in `console.log` here.

tinybike:
Note: fixes are dependent on https://github.com/AugurProject/augur.js/pull/255 being merged first.

src/modules/market/components/market-trading--confirm/market-trading--confirm.jsx
```diff
@@ -94,11 +94,12 @@ const MarketTradingConfirm = (p) => {
           onClick={(e) => {
             p.market.onSubmitPlaceTrade(p.selectedOutcome.id, (err, tradeGroupID) => {
               // onSent/onFailed CB
-              if (!err) p.toggleShowOrderPlaced()
+              if (!err) {
+                p.toggleShowOrderPlaced()
```
tinybike: Hm, is a toggle the desired behavior here?  If I submit two order tickets quickly, it doesn't seem correct to have the second order ticket submission hide the `showOrderPlaced` box.

src/modules/modal/components/modal-view/modal-view.jsx
```diff
@@ -75,10 +76,13 @@ export default class ModalView extends Component {
...
             <ModalNetworkDisconnected {...p} />
           }
+          {p.modal.type === MODAL_ACCOUNT_APPROVAL &&
+            <ModalApproval {...p} />
+          }
           {p.modal.type === MODAL_ESCAPE_HATCH &&
             <ModalEscapeHatch {...p} />
           }
-          {p.modal.canClose &&
+          {p.modal.canClose && p.modal.type !== MODAL_ACCOUNT_APPROVAL &&
```
tinybike: Is the intention for the account approval modal to be un-closeable?  I think it would be better to allow it to be closed, and just not submit the order ticket if the user closes the modal.

src/modules/trade/actions/place-trade.js
```diff
@@ -2,34 +2,68 @@ import { augur } from 'services/augurjs'
...
+          onFailed: callback,
+          onSuccess: (tradeOnChainAmountRemaining) => {
+            onComplete(tradeOnChainAmountRemaining)
+          }
+        })
+        dispatch(clearTradeInProgress(marketId))
+      }
+    }))
+  } else {
+    augur.trading.placeTrade({
```
tinybike: Ouch, I know this is just a single function call, but it's still a lot of code duplication...  Suggest defining `const placeTradeParams = { ... }` before the conditional, so that only `augur.trading.placeTrade(placeTradeParams)` has to be written out twice.

test/auth/actions/load-account-data-test.js
```diff
@@ -15,11 +15,13 @@ describe(`modules/auth/actions/load-account-data.js`, () => {
       const UpdateAssets = { updateAssets: () => {} }
       const UpdateLoginAccount = { updateLoginAccount: () => {} }
       const LoadAccountPositions = { loadAccountPositions: () => {} }
+      const approveAccounts = { checkAccountAllowance: () => {} }
```
tinybike: Suggest renaming to `ApproveAccounts` (to match the convention by the other mocks).

src/modules/modal/components/modal-view/modal-view.jsx
JohnDanz: It's actually a styling issue im side-stepping. The account approval modal is always closeable, it has a close button built into it and this was done to allow for side by side buttons for approve/close, instead of approve then a line break, close. I could remove this and just have the linebreak, but side-by-side seemed to be the better solution, so in this scenario im just ensuring that we don't have 2 close buttons. I'll discuss this quickly with @sharkcrayon to see if I shouldn't bother with side-by-side buttons and just leave it as stacked.

src/modules/trade/actions/place-trade.js
JohnDanz: good call, I agree, I think i just forgot to refactor this to be more efficient from my testing phase.

src/modules/market/components/market-trading--confirm/market-trading--confirm.jsx
JohnDanz: this should toggle extremely quickly as `onSent` usually comes back immediately, however I'll move it out to just passed the `p.prevPage` call so that it alway shows that "order placed!" ribbon once we have successfully called `placeTrade` regardless of if the `onSent` returned an error. They should still see notifications indicating a failure.

JohnDanz: oh i misunderstood, i see what you mean, it shouldn't toggle, it should force that it's shown. Gotcha. will adjust

src/modules/modal/components/modal-view/modal-view.jsx
tinybike: Ah, got it.  As long as it's closeable, fine by me.

tinybike:
Looks good, merging.

PR #1174, opened 2018-02-26
Create Market Desktop Style Updates

sharkcrayon:
https://app.clubhouse.io/augur/story/6567/create-market-view-desktop-style-updates

Just some basic CSS updates.

### DIFF
--- a/src/modules/create-market/components/create-market-form-resolution/create-market-form-resolution.styles.less
+++ b/src/modules/create-market/components/create-market-form-resolution/create-market-form-resolution.styles.less
@@ -48,12 +48,9 @@ li.CreateMarketResolution__designated-reporter-specific {
       border: none;
     }
 
-    .DateInput_input {
-      border: 2px solid transparent;
-    }
-
+    .DateInput_input_1,
     .DateInput_input__focused {
-      border-color: @color-lightpurple;
+      border: 0;
     }
 
     .DateInput__openDown_3::before,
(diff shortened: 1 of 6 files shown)


src/modules/create-market/components/create-market-form/create-market-form.styles.less
```diff
@@ -175,10 +175,10 @@
         content: "";
         display: none;
         height: 0.875rem;
-        left: 0.37rem;
+        left: 6px;
```
justinbarry: Why pixels over rem here?

src/modules/create-market/components/create-market-form-resolution/create-market-form-resolution.styles.less
```diff
@@ -42,6 +44,18 @@ li.CreateMarketResolution__designated-reporter-specific {
       z-index: 2;
     }
 
+    .SingleDatePickerInput {
```
justinbarry: Left: Seadragon
Right: Branch

Lack of contrast gives the appearance the dropdown is wider than the input box. I cannot find a mockup of this so I'm calling it out here.

screen shot 2018-02-26 at 4 08 10 pm

src/modules/create-market/components/create-market-form/create-market-form.styles.less
sharkcrayon: Because of a persistent, annoying bug. When the positioning of the inner circles of the radio buttons was set using rems instead of px, they would randomly be a little off center (sometimes, but not always, and not consistently across elements - go figure). Using px seems to have fixed it (or at least the randomness didn't occur while I was looking).

src/modules/create-market/components/create-market-form-resolution/create-market-form-resolution.styles.less
sharkcrayon: Yeah, I was torn about that too. I'm going to remove the border on focus.

PR #1177, opened 2018-02-26
7144-positions-orders-dropdown

JohnDanz:
https://app.clubhouse.io/augur/story/7144/clicking-on-open-orders-or-my-positions-title-should-expand-corresponding-section

Added the dropdown tray functionality to the positions/orders headers as well as just that little carrot on the far right.

### DIFF
--- a/src/env-dev.json
+++ b/src/env-dev.json
@@ -4,7 +4,7 @@
     "http": "http://127.0.0.1:8545",
     "ws": "ws://127.0.0.1:8546"
   },
-  "network-id": 4,
+  "network-id": 12346,
   "auto-login": true,
   "universe": null,
   "bug-bounty": false,
--- a/src/modules/market/components/market-portfolio-card/market-portfolio-card.jsx
+++ b/src/modules/market/components/market-portfolio-card/market-portfolio-card.jsx
@@ -139,7 +139,12 @@ export default class MarketPortfolioCard extends Component {
         </section>
         <section className={Styles.MarketCard__tablesection}>
           {(myPositionOutcomes || []).filter(outcome => outcome.position).length !== 0 &&
-            <div className={Styles.MarketCard__headingcontainer}>
+            <div
+              className={Styles.MarketCard__headingcontainer}
+              role="button"
+              tabIndex="0"
+              onClick={() => this.toggleTable('myPositions')}
+            >
               <h1 className={Styles.MarketCard__tableheading}>
                 My Positions
               </h1>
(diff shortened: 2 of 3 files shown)


src/env-dev.json
```diff
@@ -4,7 +4,7 @@
     "http": "http://127.0.0.1:8545",
     "ws": "ws://127.0.0.1:8546"
   },
-  "network-id": 4,
+  "network-id": 12346,
```
justinbarry: @JohnDanz Was this intentional?

JohnDanz: yeah, why are we pointing to rinkeby for local dev when we have a docker node running clique? tired of changing it on my branches, I don't believe there is any reason this has to be 4 (rinkeby).I'm trying to confirm with @stephensprinkle but he is DND today. Pretty sure this is fine, if you don't want to include it, that's fine, it's more for convenience

tinybike: I've mentioned this before, but it bears repeating: there's no reason to specify a network ID here unless it's important from a UX point of view to double-check that the ID (manually entered into this config file) matches the one the ethereum node is using.  Otherwise, the Ethereum node has only 1 network ID associated with it, so it's not an extra piece of info you have to specify.

My suggestion is just to get rid of this setting altogether, and just use network that the Ethereum node to which you're connected is on.

justinbarry: It has def gone back and forth. It would be awesome if we could drop it altogether.

tinybike: Update: we're going to get rid of this setting in [ch6692].  Until then, don't worry about what this is set to, because it's going to be different depending on each dev environment.  (Personally, I usually dev against a local Rinkeby node, so I use 4, but I know others use 12346, etc.)

src/modules/market/components/market-portfolio-card/market-portfolio-card.jsx
```diff
@@ -139,7 +139,12 @@ export default class MarketPortfolioCard extends Component {
         </section>
         <section className={Styles.MarketCard__tablesection}>
           {(myPositionOutcomes || []).filter(outcome => outcome.position).length !== 0 &&
-            <div className={Styles.MarketCard__headingcontainer}>
+            <div
+              className={Styles.MarketCard__headingcontainer}
+              role="button"
+              tabIndex="0"
```
justinbarry: The tabIndex causes a glowing blue halo to appear on click. Moving the tabindex to the svg button fixes the problem. If we are supporting tabbing we need to add a focus indicator to the svg button.

screen shot 2018-02-26 at 4 40 13 pm

JohnDanz: can't satisfy the linter and this requirement, i can just remove the halo by removing it in CSS, but it has to be tabbable if it's going to have an onclick handler for accessibility reasons

PR #1181, opened 2018-02-27
6506-outstanding-returns-styling

JohnDanz:
https://app.clubhouse.io/augur/story/6506/update-styles-for-outstanding-returns-bar-on-portfolio-my-markets

Updates the outstanding returns bar to match the updated mocks

### DIFF
--- a/src/assets/styles/_buttons.less
+++ b/src/assets/styles/_buttons.less
@@ -65,6 +65,25 @@
   cursor: default;
 }
 
+.button--offWhite {
+  .button;
+
+  background-color: @color-offwhite;
+  color: @color-purple;
+
+  &:hover {
+    background-color: @color-white;
+  }
+}
+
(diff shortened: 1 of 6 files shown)


justinbarry:
@JohnDanz Can you provide me instructions to get to the right location in the app?

JohnDanz:
Easiest way is spin up a local environment, use the account that loads all the canned data and you should be able to navigate to my Markets. You won't see outstanding returns because there are none to collect yet, so go to `src/modules/market/components/market-preview/market-preview.jsx` line 18 and remove the `&& p.unclaimedCreatorFees.value > 0 ` on that conditional for a moment. you should see a rendered outstanding returns bar now under the market cards with 0 ETH to claim.

justinbarry:
Looks good!

screen shot 2018-03-01 at 8 32 09 am

PR #1182, opened 2018-02-27
additional details fix

bthaile:
found issue where `additional details` button doesn't show b/c there isn't any extra info but there is always the case of showing resolution source. fixed to always show `additional details` button on report form.


justinbarry:
Confirmed.
screen shot 2018-02-27 at 1 56 41 pm

PR #1184, opened 2018-02-27
7239-update-reconnection-modal

JohnDanz:
https://app.clubhouse.io/augur/story/7239/distinguish-between-augur-and-eth-node-disconnected-on-reconnecting-modal

So for this update the best way to test it is to run your own local ethereum node/augur-node setup. 

Once you are fully spun up, connected, and logged into an account, kill your augur-node. You should be prompted with a modal, and if you open the console you should see a number of calls to connectAugur, once every 3 seconds or so. If you turn your augurnode back on, it should reconnect automatically and bring you back to the categories home page. You can do the same for the ethereum node, but i recommend saving that test for the end because you will need to reupload contracts and all that again if you want to continue after looking at it.

The text should change based on if you are not connected to the Augur Node, the Ethereum Node, or Both. If both the text is shorter and more generic.

When you click on the link for `here` you should see a form appear. Once this form is visible, the reconnection should be paused. if you check `state.connection.isReconnectionPaused` it should be `true` while the form is open. Once you submit the form or click on the `here` link again, `state.connection.isReconnectionPaused` should be `false` and you should see augur attempting to reconnect again in the console.


src/modules/app/actions/update-connection.js
```diff
@@ -22,3 +23,14 @@ export function updateAugurNodeConnectionStatus(isConnected) {
     isConnected
   }
 }
+
+/**
+ * @param {Boolean} isReconnectionPaused
+ * @return {{type: string, isReconnectionPaused: *}} returns action
+ */
+export function updateIsReconnectionPaused(isReconnectionPaused) {
```
justinbarry: Please add test coverage.

src/modules/app/reducers/connection.js
```diff
@@ -15,6 +15,11 @@ export default function (connection = DEFAULT_STATE, action) {
         ...connection,
         isConnectedToAugurNode: action.isConnected
       }
+    case UPDATE_IS_RECONNECTION_PAUSED:
```
justinbarry: Please add test coverage.

src/modules/app/actions/listen-to-updates.js
```diff
@@ -195,13 +196,19 @@ export function listenToUpdates(history) {
 
       const retry = (cb) => {
         const { connection, env } = getState()
-        if (!connection.isConnected) {
+        if (!connection.isConnected || !connection.isConnectedToAugurNode) {
```
justinbarry: Please add test coverage.

justinbarry:
Added this to address missing connections on initial load.
https://app.clubhouse.io/augur/story/7334/define-behavior-for-app-when-augur-node-is-not-available-on-initial-load

justinbarry:
For whatever reason it leaves me logged out after the connection restores.

screen shot 2018-03-01 at 4 45 52 pm

PR #1185, opened 2018-02-27
7048 gas est reporting

bthaile:
create separate call to get gas estimates to display to initial reporter. convert gas to gas costs using constants, round to 4 decimals based on UI mocks

https://app.clubhouse.io/augur/story/7048/add-gas-costs-to-reporting-report-confirm-form


src/modules/reporting/actions/submit-initial-report.js
```diff
@@ -28,3 +29,29 @@ export const submitInitialReport = (marketId, selectedOutcome, invalid, history,
     })
   }
 }
+
+
+export const estimateSubmitInitialReport = (marketId, callback = logError) => (dispatch, getState) => {
```
justinbarry: Please add test coverage

src/modules/reporting/components/reporting-report/reporting-report.jsx
```diff
@@ -88,8 +92,21 @@ export default class ReportingReport extends Component {
         reporterGasCost: formatEtherEstimate(marketCreationCostBreakdown.targetReporterGasCosts),
         stake: this.props.isOpenReporting ? '0' : repAmount.formatted,
       })
+    })
+  }
+
+  calculateGasEstimates() {
```
justinbarry: Please breakout out into a util file and add test coverage.

src/modules/reporting/actions/submit-initial-report.js
bthaile: It simple server call no logic to test

src/modules/reporting/components/reporting-report/reporting-report.jsx
bthaile: ok, will break out.

src/modules/reporting/actions/submit-initial-report.js
bthaile: ok, will write unit test for logic based on service call success or fails. makes sense

bthaile: done unit tests written

bthaile:
@justinbarry 
are you suggesting breaking out the `calculateGasEstimates` method out of the component. I broke out formatting gas cost to ether into a help method with unit tests.

justinbarry:
@bthaile Maybe? Breaking out the formatter like you did is good enough for me. 

If we have a ban on  async inside components then maybe it should move but this is something we should decide as a team. Balancing the practical with the fancy is hard sometimes ;)

bthaile:
resolved merge conflict should be good to go.

justinbarry:
Looks good!

screen shot 2018-03-02 at 12 59 18 am

PR #1187, opened 2018-02-28
Added Shapeshift buttons for depositing ETH & REP to user's account

adrake33:


src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -23,6 +30,16 @@ export default class AccountDeposit extends Component {
       height: 'auto',
       width: '100%',
     }
+    const shapeShiftButton = {
```
justinbarry: Please move to less style sheet.

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -13,6 +13,13 @@ export default class AccountDeposit extends Component {
     address: PropTypes.string.isRequired
   }
 
+  static shapeShiftOnClick(e) {
+    e.preventDefault()
```
justinbarry: ` e.preventDefault()` and `return false` are more or less redundant.

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -13,6 +13,13 @@ export default class AccountDeposit extends Component {
     address: PropTypes.string.isRequired
   }
 
+  static shapeShiftOnClick(e) {
```
justinbarry: At least in my local testing, this guy does not need to be static. Try:

```<button onClick={e => this.shapeShiftOnClick(e)}  ...></button>```

File: src/modules/account/components/account-deposit/account-deposit.jsx [hunk 4]
```diff
@@ -23,6 +30,16 @@ export default class AccountDeposit extends Component {
...
     }
+    const shapeShiftButton = {
+      clear: 'both',
+      float: 'left',
+      fontWeight: 'bold',
+      textTransform: 'uppercase',
+    }
+    let shapeShiftConverter = <a href="https://shapeshift.io">Use Shapeshift</a>
+    if (parseInt(window.augur.rpc.getNetworkID(), 10) === 1) {
+      shapeShiftConverter = <div><button onClick={(e) => { AccountDeposit.shapeShiftOnClick(e) }} value={'https://shapeshift.io/shifty.html?destination=' + p.address + '&output=ETH'} style={shapeShiftButton}>ShapeShift to ETH</button><button onClick={(e) => { AccountDeposit.shapeShiftOnClick(e) }} value={'https://shapeshift.io/shifty.html?destination=' + p.address + '&output=REP'} style={shapeShiftButton}>ShapeShift to REP</button></div>
```
justinbarry: This can be broken apart into multiple lines for legibility.

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -23,6 +30,16 @@ export default class AccountDeposit extends Component {
       height: 'auto',
       width: '100%',
     }
+    const shapeShiftButton = {
+      clear: 'both',
+      float: 'left',
+      fontWeight: 'bold',
```
justinbarry: Use less variable here when it is moved over.

src/modules/account/components/account-deposit/account-deposit.jsx
justinbarry: Or if you want to get really hardcore:

```
  shapeShiftOnClick(address, output) {
    return (e) => {
      e.preventDefault()
      const link = `https://shapeshift.io/shifty.html?destination=${address}&output=${output}`
      window.open(link, '1418115287605', 'width=700,height=500,toolbar=0,menubar=0,location=0,status=1,scrollbars=1,resizable=0,left=0,top=0')
    }
  }
```

And the onClick handlers would look like this:

```
onClick={this.shapeShiftOnClick(p.address, 'ETH')}
```

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -23,6 +30,16 @@ export default class AccountDeposit extends Component {
...
+      textTransform: 'uppercase',
+    }
+    let shapeShiftConverter = <a href="https://shapeshift.io">Use Shapeshift</a>
+    if (parseInt(window.augur.rpc.getNetworkID(), 10) === 1) {
```
justinbarry: No need for window here. Add the following to import it.

```
import { augur } from 'services/augurjs'
```

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Oops, these were in the original JS provided by ShapeShift, but since I changed the anchor tags to buttons, I suppose neither is needed now. Will remove.

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Sorry, I'm pretty new to the UI code. Which CSS file should this be moved to?

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Ah, ok. The linter was complaining when I originally tried `augur.rpc.getNetworkID()`, and I wasn't sure how to resolve it. I'll add that import line.

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: I'm not really familiar with less. What is the variable I should be using?

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Never mind, looks like it's `account-deposit.styles.less`. I'll add it there.

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: I tried both of these approaches, but the linter still gives me the following error when I try to push my changes:
```
/Users/aaron/Documents/Augur/augur/src/modules/account/components/account-deposit/account-deposit.jsx
  21:20  error  Expected 'this' to be used by class method 'shapeShiftOnClick'  class-methods-use-this

✖ 1 problem (1 error, 0 warnings)
```

src/modules/account/components/account-deposit/account-deposit.jsx
justinbarry: I think you found it but if not it should be the less file in the same directory.

src/modules/account/components/account-deposit/account-deposit.jsx
justinbarry: We can screenshare if that would be helpful.

src/modules/account/components/account-deposit/account-deposit.jsx
justinbarry: Awesome!

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Sure, let's look into this more after today's stand-up meeting.

adrake33:
@justinbarry Can you please review these updates and let me know if everything looks ok?

justinbarry:
Looks good!

FYI, @adrake33 I updated the styles to match the "USE SHAPESHIFT" copy.

screen shot 2018-03-02 at 11 15 09 am

screen shot 2018-03-02 at 9 50 16 am

PR #1189, opened 2018-03-01
Revised text on Withdraw page

adrake33:
Changed text as per https://app.clubhouse.io/augur/story/7099/review-and-correct-account-withdraw-text.


adrake33:
Sprinkle confirmed on https://app.clubhouse.io/augur/story/7099/review-and-correct-account-withdraw-text that this new text is correct.

PR #1191, opened 2018-03-01
4627-update-error-page

JohnDanz:
https://app.clubhouse.io/augur/story/4627/update-error-boundary-page-styling

This will remove the (ノಠ益ಠ)ノ彡┻━┻ guy forever. may he RIP in peace. ;)

Best way to test this is to go to the `src/modules/market/components/market-view/market-view.jsx` component and add something like this `{p.notARealValue.length}` to the render method somewhere. Now fire up the UI, and click the `trade` button on any market card and you should get an exception which should present to you the new error splash screen with the text from the mock and a button you can use that should send you back to the home page and refresh.


justinbarry:
@JohnDanz Do we have mocks for this screen?

RIP flippy tabble throwy guy.

screen shot 2018-03-02 at 12 01 38 pm

JohnDanz:
We only had the screenshot on the CH issue, which i don't believe is mocked. I confirmed with @sharkcrayon that this is ok for now and we can change the styling later if we get better mocks/add the network background thing.

PR #1192, opened 2018-03-01
comma-dangle for consistency

stephensprinkle:
Noticed we were pretty inconsistent w/ comma-dangle formatting since the rule was disabled.

This enforces a trailing comma + gets us in-line w/ the augur.js rule.

Feel free to close if this is not the route we want to go.

(Changes are all autogen’d via `—fix`)


justinbarry:
@stephensprinkle I'm all for this!

PR #1193, opened 2018-03-01
5884 upcoming dispute

bthaile:
followed existing pattern for loading markets that are about to be or in crowd dispute. This is only the back end loading portion.


src/modules/reporting/actions/load-disputing.js
```diff
@@ -0,0 +1,47 @@
...
+
+  augur.augurNode.submitRequest(
+    'getMarkets',
+    {
+      reportingState: constants.REPORTING_STATE.AWAITING_NEXT_WINDOW,
+      sortBy: 'endDate',
+      ...args,
+    },
+    (err, result) => {
+      if (err) return callback(err)
```
justinbarry: Missing test coverage of this conditional.

PR #1238, opened 2018-03-11
6362 dispute progress

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/6362)

to Test:

after designate-report on market start disputing with dispute-contribute. When a dispute happens the progress bar should increase and the needed REP to fill bond value decreases in the dispute form.

From the reporting-dispute-market page there should be a dispute button on markets that have been initially reported on.

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 

### DIFF
--- a/src/modules/market/containers/market-properties.js
+++ b/src/modules/market/containers/market-properties.js
@@ -9,7 +9,7 @@ const mapStateToProps = (state, ownProps) => ({
   isLogged: state.isLogged,
   isMobile: state.isMobile,
   loginAccount: state.loginAccount,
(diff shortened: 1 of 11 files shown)


bthaile:
I noticed this, we need to get on the same page. I talked it over with Scott.
The DisputeRound data coming from AN indicates the round that has been completed. Therefore in the reporting dispute markets page is shows the round finished/completed. I increment the value indicating the current active round.

I'll make the change, best we look consistent, really it doesn't matter.

PR #1250, opened 2018-03-13
7877-min-max-scalar-values

JohnDanz:
fixes the scalar ...scales... on the market cards so that they don't always show 0%-100% but instead use the min/max of the scalar market. Removed the hardcoded % denominations from scalar, will replace with real denominations when AugurNode is updated to return them to the UI.

[Clubhouse Story](https://app.clubhouse.io/augur/story/7877/min-and-max-are-no-longer-displaying-for-scalar-markets)

## Submitter checklist
  - [x] Steps to reproduce
  - [x] Screenshot (If applicable)

## Reviewer Checklist
- [x] Test/lint pass 
- [x] Post merge, story marked complete

### DIFF
--- a/src/modules/market/components/market-basics/market-basics.jsx
+++ b/src/modules/market/components/market-basics/market-basics.jsx
@@ -69,7 +69,7 @@ const MarketBasics = (p) => {
         </h1>
 
         {(p.marketType === BINARY || p.marketType === SCALAR) &&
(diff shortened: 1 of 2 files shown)


JohnDanz:
steps to reproduce, look at any scalar market card in the markets page and it should now correctly show the min/max and no % by those values. 

Example a -10  to 120 ranged scalar would show:

`-10     55      120`

PR #1255, opened 2018-03-14
7637-sell-order-direction

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/7637/sell-order-positions-are-not-displayed-as-negative)

fixed it so that if you have an open sell order the value will be shown as `-N` where N is the num of shares. Previously it was always showing `+N` so it was difficult to tell between open buy or sell orders.

## Submitter checklist
  - [x] Steps to reproduce
  - [x] Screenshot (If applicable)

## Reviewer Checklist
- [x] Test/lint pass 


src/modules/market/components/market-positions-list--order/market-positions-list--order.jsx
```diff
@@ -66,7 +66,7 @@ export default class Order extends Component {
           }
         </li>
         <li>
-          { p.type === 'sell' ? <span>-</span> : <span>+</span> } { getValue(p, 'order.unmatchedShares.formatted') }
```
justinbarry: Do we have 'sell' as a constant someplace?

PR #1279, opened 2018-03-21
7950-reporting-time-graph

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/7950/adjust-reporting-time-remaining-pie-graph)

you will need to use flash to test this, and also will need to absolute value the `percentageElapsed` calculation inside of the `src/modules/markets/components/common/time-remaining-indicator/time-remaining-indicator.jsx` file. This is because `Date.now()` uses system time, but the markets will have been "flashed" forward. So you will always be calculated as before the reporting time because the system clock is before the right time. 

use the following commands on a market:
`node scripts/flash designate-report <MARKET_ID>,<OUTCOME_ID>` notice that there is no space around that comma for the args.
Go to the reporting page on the main account and you should see these markets with the circle in the top right corner of the market cards.

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 


justinbarry:
@JohnDanz time-remaining-indicator.jsx moves to use blocktime here #1276.

PR #1288, opened 2018-03-22
7119-fix-empty-categories

JohnDanz:
prevented the UI from selecting empty categories, also disabled empty category names from being displayed if they somehow get through, just to be extra sure. Updated unit tests

[Clubhouse Story](https://app.clubhouse.io/augur/story/7119/empty-category-on-categories-page)

To reproduce, change the default state for the categories reducer to look like so:
`const DEFAULT_STATE = { 99: { popularity: 0, category: '' } }` 

this will force the state to have a bunk category in it. load up the UI, that bunk category should be in the state, yet not be counted or shown in the categories view.

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 


src/modules/categories/selectors/categories.js
```diff
@@ -8,9 +8,20 @@ export default function () {
 
 export const selectCategories = createSelector(
   selectCategoriesState,
-  categories => Object.keys(categories || {})
-    .map(category => ({ popularity: categories[category].popularity, category: categories[category].category }))
-    .sort(popularityDifference),
+  (categories) => {
```
justinbarry: FWIW this can be rewritten as
```
export const selectCategories = createSelector(
  selectCategoriesState,
  categories =>
    Object.values(categories)
      .filter(c => c && !!c.category)
      .map(({ category, popularity }) => ({ category, popularity }))
      .sort(popularityDifference),
)
```

src/modules/categories/components/category-list/category-list.jsx
```diff
@@ -19,11 +19,11 @@ const CategoryList = (p) => {
         return (
           <div
             className={categoryStyling}
-            key={category !== null ? JSON.stringify(category) : `${JSON.stringify(category)}${categoryIndex}`}
+            key={category ? JSON.stringify(category) : `${JSON.stringify(category)}${categoryIndex}`}
           >
             <Category
-              category={category !== null ? category.category : 'null-category'}
-              popularity={category !== null ? category.popularity : 0}
+              category={(category && !!category.category) ? category.category : 'null-category'}
```
justinbarry: Isn't this already handled in the selector?

test/categories/selectors/categories-test.js
```diff
@@ -137,4 +137,32 @@ describe(`modules/categories/selectors/categories.js`, () => {
       ])
     },
   })
+  test({
```
justinbarry: Sweet.

src/modules/categories/components/category-list/category-list.jsx
JohnDanz: yeah, just making absolutely sure. Bunk/bad categories can still be in the state, they shouldn't be, but can be. the selector should be filtering them out before getting here, but in the off chance that somehow it didn't for whatever reason, i wanted to be explicit.

PR #1291, opened 2018-03-22
Update kickoff script

justinbarry:
Let me know what you think.

### DIFF
--- a/scripts/setup-local-dev.sh
+++ b/scripts/setup-local-dev.sh
@@ -15,7 +15,7 @@ export ETHEREUM_PRIVATE_KEY=${ETHEREUM_PRIVATE_KEY:-fae42052f82bed612a724fec3632
 
 CONTAINER_NAME="geth-node";
 if $USE_POPULATED_CONTRACTS; then
-  DOCKER_IMAGE="augurproject/dev-pop-geth:core-$(npm view augur-core version)";
+  DOCKER_IMAGE="augurproject/dev-pop-geth:latest";
 else
   DOCKER_IMAGE="augurproject/dev-node-geth:latest";
 fi
--- a/scripts/setup-local-dev.sh
+++ b/scripts/setup-local-dev.sh
@@ -11,18 +11,17 @@ export ETHEREUM_WS=http://127.0.0.1:8546;
 
 export ETHEREUM_PRIVATE_KEY="fae42052f82bed612a724fec3632f325f377120592c75bb78adfcceae6470c5a";
 
-SHOULD_SETUP_MARKETS=false;
+CONTAINER_NAME="geth-node";
+DOCKER_IMAGE="augurproject/dev-pop-geth:core-$(npm view augur-core version)";
 
-set +e;
-GETH_STATE=$(docker inspect -f {{.State.Running}} geth-node);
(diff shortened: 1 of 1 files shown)


scripts/setup-local-dev.sh
```diff
@@ -11,18 +11,17 @@ export ETHEREUM_WS=http://127.0.0.1:8546;
...
-set +e;
-GETH_STATE=$(docker inspect -f {{.State.Running}} geth-node);
-if [[ "$GETH_STATE" != "true" ]]; then
-  echo "geth-node: ";
-  docker run -it -d --rm --name geth-node -p 8545:8545 -p 8546:8546 augurproject/dev-node-geth:latest;
-  SHOULD_SETUP_MARKETS=true;
-else
-  echo "geth-node: already running";
+CONTAINER_NAME="geth-node";
+DOCKER_IMAGE="augurproject/dev-pop-geth:core-$(npm view augur-core version)";
```
bthaile: dev-pop-geth:latest would work. There might be a chance that augur.js updates to use augur-core version that doesn't have corresponding docker image created. possible since creating the pop docker image is a manual process. still recommend :latest

scripts/setup-local-dev.sh
```diff
@@ -11,18 +11,17 @@ export ETHEREUM_WS=http://127.0.0.1:8546;
 
 export ETHEREUM_PRIVATE_KEY="fae42052f82bed612a724fec3632f325f377120592c75bb78adfcceae6470c5a";
```
bthaile: env. variables are no longer needed since deploy is already done

scripts/setup-local-dev.sh
```diff
@@ -32,7 +31,8 @@ rm -rf node_modules;
 yarn;
 yarn run build;
 yarn link;
-$SHOULD_SETUP_MARKETS && yarn run deploy:environment;
+
+docker cp $CONTAINER_NAME:/augur.js/src/contracts/addresses.json src/contracts/addresses.json;
```
bthaile: I guess nice to have (correct contracts) but not needed if using :latest, just need latest augur.js linked in to augur-node and augur ui.

scripts/setup-local-dev.sh
justinbarry: Even for augur-node?

scripts/setup-local-dev.sh
justinbarry: Same reason as above. I want to support working on an old branch.

scripts/setup-local-dev.sh
justinbarry: I did this to deal with old branches. Unsure how often this will happen.

scripts/setup-local-dev.sh
bthaile: augur-node doesn't need the env variables, flash uses them

epheph:
@justinbarry @bthaile i added support for a new variable. By default, this uses pre-populated contracts, but if you set `USE_POPULATED_CONTRACTS=false`, you get the previous behavior (which is useful for iterating on new market data or testing overall deployment locally

epheph:
(and the re-introduction of the old way means the variables are now worthwhile again)

PR #1292, opened 2018-03-22
simply removed 'No' from binary market liquidity in create market.

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/7217)

super simple remove 'No' from create binary market liquidity page. repro on ch task.

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [x] Test/lint pass 
- [x] Post merge, story marked complete


tinybike:
This is an improvement, but I think it would be better for binary markets' add liquidity interface to look like the one for scalars; i.e., not have an outcome label at all.

bthaile:
https://app.clubhouse.io/augur/story/7217/binary-order-form-in-create-market-should-show-only-yes-value

Updated ch task. What do you think about making binary market trade look the same for new-market liquidity and for binary market trade?

bthaile:
I can remove the outcome label, I have a question out to @sharkcrayon in ch task 7217, might be a little bigger than this PR.

tinybike:
Correct @bthaile -- in both cases, I think that the binary order ticket should not have a radio button for the outcome or display the outcome label in the order ticket.  (Note: getting rid of the radio button for binaries is actually part of the existing task, although that hasn't been done in this PR yet.)

bthaile:
updated PR removed 'Yes' for create binary market liquidity.

src/modules/create-market/components/create-market-form-liquidity/create-market-form-liquidity.jsx
```diff
@@ -45,9 +45,10 @@ export default class CreateMarketLiquidity extends Component {
       orderEstimate: '',
       minPrice: new BigNumber(0),
       maxPrice: new BigNumber(1),
-      selectedOutcome: this.props.newMarket.type === SCALAR ? 0 : '',
       selectedNav: BID,
     }
+    this.state.selectedOutcome = this.props.newMarket.type === SCALAR ? 0 : ''
```
tinybike: This should be `1` for both scalars and binaries.

bthaile: oh, this was existing code. I'll make the change.

bthaile:
updated PR to default to 1 for both binary and scalar markets.

PR #1293, opened 2018-03-22
augur-node changed property need to fix in UI.

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/5906)

description on ch task

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 


tinybike:
I'm not entirely clear what this PR is for.  The description on CH is just:
```
What are the display conditions for Position Close button.
If position has already been closed?
If 0 shares on position?
Need input from @serena
```

bthaile:
I noticed when Sprinkle did a trading demo and closed an open order. the order 'pending cancellation' didn't get cleared. I found that augur-node changed a property UI needed in order to clear the order. 

The comment about 0 shares is unrelated to the code changes. Not sure why open orders would show 0 shares.

src/modules/transactions/actions/construct-transaction.js
```diff
@@ -139,7 +139,10 @@ export function constructTransferTransaction(log, address) {
 }
 
 export const constructCancelOrderTransaction = (trade, marketId, marketType, description, outcomeId, outcomeName, minPrice, maxPrice, status) => (dispatch, getState) => {
-  const displayPrice = augur.trading.denormalizePrice({ normalizedPrice: trade.price, minPrice, maxPrice })
+  const { orderBooks } = getState()
+  const type = trade.orderType === '0' ? TYPES.BUY : TYPES.SELL
+  const price = !trade.price ? orderBooks[marketId][outcomeId][type][trade.orderId].price : trade.price
```
tinybike: I'm not sure I understand what's going on here, is `price` not always included with the trade?  If not, suggest just including it with augur-node's payload.

Note that this code is all significantly modified on the rollback branch, so it's preferable not to also make changes to it here.

src/modules/app/actions/listen-to-updates.js
```diff
@@ -100,12 +100,12 @@ export function listenToUpdates(history) {
           } else {
             console.log('OrderCanceled:', log)
             // if this is the user's order, then add it to the transaction display
-            if (log.orderCreator === getState().loginAccount.address) {
+            if (log.sender === getState().loginAccount.address) {
```
tinybike: Is `orderCreator` no longer included with the log?

bthaile: changed to sender from what I see with the newest version of augur-node.

src/modules/transactions/actions/construct-transaction.js
bthaile: ok, then only changing listen-to-update with property change and re-ordering method calls will fix the issue.

bthaile:
corrections made, only change is to listen-to-updates

tinybike:
LGTM

PR #1310, opened 2018-03-27
4971-connection-modal

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/4971/connect-to-augur-modal)

separated out the connection modal from the reconnection modal. Connection modal should now appear if you fail to connect to the assigned env values or if your env is blank. Priority for connections to geth node is as follows:

Metamask (web3) > IPC Address > WS Address > HTTP Address. 

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 


src/modules/modal/components/modal-network-connect/modal-network-connect.jsx
```diff
@@ -0,0 +1,215 @@
...
+        ...ethNode,
+      },
+    }
+    p.updateEnv(updatedEnv)
+    // p.submitForm used as a hook for disconnection modal, normally just preventsDefault
+    p.submitForm(e)
+    // reset local error state and initial attemptConnection loading icon
+    this.setState({ isAttemptingConnection: true, connectErrors: [] })
+
+    p.connectAugur(p.history, updatedEnv, !!p.modal.isInitialConnection, (err, res) => {
```
justinbarry: Missing in prop-types

src/modules/modal/components/modal-network-connect/modal-network-connect.jsx
```diff
@@ -0,0 +1,215 @@
...
+      } else if (props.env['ethereum-node'].http) {
+        ethereumNode = props.env['ethereum-node'].http
+      }
+    }
+
+    this.state = {
+      augurNode: props.env['augur-node'] || '',
+      ethereumNode,
+      isAttemptingConnection: false,
+      isWeb3Available: (!!window && !!window.web3),
```
justinbarry: Should we be polling for this?

JohnDanz: I don't think that's needed, basically I only need to care if it's there on this form so we can hide the input for ethereumNode because you are already connected through metamask. I can't think of a reason we would poll for it over just checking if it's there or not when we go to construct this component.

PR #1311, opened 2018-03-28
Bump augur.js

pgebheim:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,11 +114,6 @@
         }
       }
     },
-    "@types/bignumber.js": {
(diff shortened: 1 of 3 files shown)


pgebheim:
This needs to be updated with the real yarn.lock -- I can't install nodehid with yarn so I can't update that.

Also, this should work out of the box with dev-pop-node:latest

Bring up a local node with
```
docker pull augurproject/dev-pop-node:latest
docker run -e GETH_VERBOSITY=4 -it -p 8545:8545 -p 8546:8546 augurproject/dev-pop-geth:latest
```

PR #1315, opened 2018-03-29
6698-filter-markets-tag

JohnDanz:
fixed an issue with tags not filtering when selected on the side menu, fixed another visual issue where tags were displaying for markets not present in the filtered view.

[Clubhouse Story](https://app.clubhouse.io/augur/story/6698/unable-to-filter-market-categories-by-tag)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass 

### DIFF
--- a/src/modules/filter-sort/helpers/filter-by-tags.js
+++ b/src/modules/filter-sort/helpers/filter-by-tags.js
@@ -6,9 +6,9 @@ export default function filterByTags(tags, items) {
   return items.reduce((p, item, i) => {
     if (tags.every(filterTag =>
       item.tags.some((tag, tagIndex) =>
-        tagIndex !== 0 && tag === filterTag))
+        tag === filterTag))
     ) {
-      return [...p, i]
+      return [...p, items[i].id]
     }
 
     return p
--- a/src/modules/markets/components/markets-view.jsx
+++ b/src/modules/markets/components/markets-view.jsx
@@ -13,7 +13,9 @@ import isEqual from 'lodash/isEqual'
 // import makePath from 'modules/routes/helpers/make-path'
 
 // import { FAVORITES, MARKETS } from 'modules/routes/constants/views'
-import { CATEGORY_PARAM_NAME, FILTER_SEARCH_PARAM } from 'modules/filter-sort/constants/param-names'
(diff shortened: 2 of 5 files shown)


src/modules/filter-sort/helpers/filter-by-tags.js
```diff
@@ -6,9 +6,9 @@ export default function filterByTags(tags, items) {
   return items.reduce((p, item, i) => {
     if (tags.every(filterTag =>
       item.tags.some((tag, tagIndex) =>
-        tagIndex !== 0 && tag === filterTag))
+        tag === filterTag))
     ) {
-      return [...p, i]
+      return [...p, items[i].id]
```
stephensprinkle: This helper function should be returning the index of the item, not the id of the item as other components (namely the rest of the filter-sort module) is architected with the expectation that this just returns an array of matched indexes.

src/modules/markets/components/markets-view.jsx
```diff
@@ -66,6 +68,14 @@ export default class MarketsView extends Component {
   }
 
   componentWillUpdate(nextProps, nextState) {
+    const newTags = decodeURIComponent(parseQuery(nextProps.location.search)[TAGS_PARAM_NAME] || '')
```
stephensprinkle: No need for the `decodeURIComponent` here.

Also, I’d do a conditional around just `location.search` as parsing that is a performance hit on each update.

src/modules/markets/components/markets-view.jsx
```diff
@@ -66,6 +68,14 @@ export default class MarketsView extends Component {
...
+    if (newTags.length) {
+      const marketsFilteredByTags = filterByTags(newTags.split(QUERY_VALUE_DELIMITER), nextProps.markets)
```
stephensprinkle: Per my comments on `filterByTags`, you’ll need to update handling of this value accordingly.

src/modules/markets/components/markets-view.jsx
JohnDanz: i was finding there was a need as tags such as `US Politics` where coming in URI encoded like so `US%20Politics`

stephensprinkle:
Per discussion, merging down as is as the utilization of the filter-sort module will require a broader refactor.

PR #1335, opened 2018-03-30
Fix kickoff

pgebheim:


bthaile:
Need description for everything this PR is doing. Looks to be a bit more then fixing kickoff script.

bthaile:
IIRC, you wanted to bump UI augur.js version with augur-node version.

PR #1355, opened 2018-04-03
Edge login

swansontec:
[Clubhouse Story](https://app.clubhouse.io/augur/story/6044)

This replaces the old Airbitz login system (which is currently disabled) with the new Edge-branded login system.

The main API difference I had to account for in this pull request is that the Edge login system's `makeEdgeUiContext` is asynchronous, unlike the Airbitz one, since it needs to communicate with the iframe before it can do anything. Therefore, I create the Edge context in the background the first time either button appears, and put it into redux when it finishes. Once it becomes available, the Edge login button goes from disabled to enabled.

I would like to update the Airbitz logo to an Edge logo, but I'm not sure what your workflow is for getting SVG's converted into JSX. I have the assets in SVG format already, so I can update the PR.

The Edge iframe contents are currently being served from the `assets/` folder. Ideally, they would be served from a different domain than the main Augur app. This will create a security boundary between the Edge login system and the Augur app. Right now, the security boundary isn't as hard as it could / should be. 

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [x] Test/lint pass 


package.json
```diff
@@ -40,6 +40,7 @@
     "clipboard": "1.7.1",
     "core-js": "2.5.1",
     "d3": "4.12.0",
+    "edge-login-ui-web": "^0.2.5",
```
tinybike: Version number should be fixed (no caret).

tinybike:
@swansontec I tried to use "connect my edge account" twice, logged in by scanning the QR code, and it created a new account each time.  Shouldn't this use the existing account / how do I do that?

package.json
tinybike: We do use lock files (yarn.lock and package-lock.json).

swansontec:
Aha! Found the problem. It's a timing issue in our cross-domain iframe logic. I'll get it fixed in the `edge-login-ui-web` library and make a new release.

swansontec:
I fixed the edge-login-ui-web module to properly report which user accounts are available on the device. This was the race condition I was mentioning.

However, after re-reading your comment, I'm not exactly sure what you are describing. Can you clarify the issue you are seeing?

tinybike:
@swansontec Logging in with a password fails for me with the following error:
```
starting scrypt n=16384 r=1 p=1
c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484 finished scrypt n=16384 r=1 p=1 in 236.49999999997817ms
c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484 starting scrypt n=16384 r=1 p=1
c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484 Context.loginWithPassword TypeError: g.push is not a function
    at e (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at t (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at h (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484
    at new Promise (<anonymous>)
    at Object.dt [as scrypt] (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
    at Object.<anonymous> (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
    at o (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484 Context.loginWithPassword TypeError: g.push is not a function
    at e (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at t (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at h (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:45)
    at c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484
    at new Promise (<anonymous>)
    at Object.dt [as scrypt] (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
    at Object.<anonymous> (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
    at o (c1742f9ed7cee5d6bd6e0c9c5ad4a525.js:484)
```
However, I tried two other accounts and password login worked fine for both of them.  The failing account's password is much longer than the others; possibly that makes a difference?

tinybike:
@swansontec I also just noticed that the "Setup Recovery" button doesn't actually have you set up recovery, instead it just closes the modal and sends you to the categories page.

paullinator:
@tinybike The setup recovery bug actually got found and fixed in our code but missed this PR. We have new one coming in a day or two to fix this and a few other UI issues.

swansontec:
@tinybike Looks like the login failure you are seeing is indeed related to having a really long password. If the password is over 64 characters, the scrypt library we are using crashes. This is because the scrypt library expects to call `Array.push` inside the sha256 routine, but we are passing in a `Uint8Array`, which doesn't have that method.

I have made the fix in our core library (https://github.com/Airbitz/edge-core-js/pull/68), and once it gets in, I'll update this pull request with the latest version.

tinybike:
Ah ok @swansontec -- the password I was using was exactly 64 characters.  Will try this out again once the PR is updated!

swansontec:
This upgrade should resolve the long-password issue you mentioned. It also includes fixes a bunch of other little UI stuff.

tinybike:
@swansontec I get the following error in the browser console (Chrome) when I build and run the UI locally, navigate to accounts, connect an account, then try to click on "Connect Edge Account" (which is greyed out):
```
post-robot.js?6573:2594 [post-robot] fullpage :: localhost:8080/ 

     #send response :: postrobot_ready :: http://localhost:8080 

 {target: undefined, hash: "postrobot_ready_33efed7ab5", name: "postrobot_ready", type: "postrobot_message_response", ack: "error", …}
```
Does this look like something on your end?  I've never seen this error before.

tinybike:
I also see this error in the browser console:
```
post-robot.js?6573:2658 Uncaught Error: Window did not load after 15000ms
    at eval (post-robot.js?6573:2658)
```
FWIW I'm currently on a relatively slow connection.  I will try again later on a faster connection in case it's just that.

swansontec:
We use a library called post-robot to handle the cross-domain iframe messaging. Both errors you reported are related to that library.

In the first case, it shouldn't be possible to click the button while it's greyed out - the iframe hasn't loaded, so we aren't ready to show the user anything. Once the button turns blue, everything should still work, I'm assuming. Otherwise, I might need help reproducing the problem.

In the second case, the error is probably related to a slow connection. If we can't load the iframe in 15 seconds, the post-robot library times out. We may want to raise the timeout too, or start the loading process earlier (when Augur loads, as opposed to when the user first visits the Edge login component).

tinybike:
@swansontec I just tried again, this time on a fast connection, and I'm still getting the same errors:
```
post-robot.js?6573:2594 [post-robot] fullpage :: localhost:8080/ 

	 #send response :: postrobot_ready :: http://localhost:8080 

 {target: undefined, hash: "postrobot_ready_c40f078312", name: "postrobot_ready", type: "postrobot_message_response", ack: "error", …}

post-robot.js?6573:2658 Uncaught Error: Window did not load after 15000ms
    at eval (post-robot.js?6573:2658)
```
I realize the button isn't clickable when it's greyed out.  These errors appear in the console without me clicking on the button at all.  The button never turns blue.

swansontec:
I see. Can you give me some insight into what mean by "build and run the UI locally"? If I clone the augur/edge-login branch and run `yarn && yarn dev`, things seem to work fine, but it sounds like you are doing something different here.

If you clone the `edge-login-ui` repo and run `yarn`, you should find the iframe assets in `packages/edge-login-ui-web/assets`. If you copy this into your own `assets` folder, it should work along with the bridge files located in `packages/edge-login-ui-web/lib`.

tinybike:
@swansontec It sounds like we're doing the same thing; I'm also just running `yarn && yarn dev` on your branch.

> If you clone the edge-login-ui repo and run yarn, you should find the iframe assets in packages/edge-login-ui-web/assets. If you copy this into your own assets folder, it should work along with the bridge files located in packages/edge-login-ui-web/lib.

Not sure I'm following.
1. When you say to copy the files in `packages/edge-login-ui-web/assets` to `augur/src/assets`, do you mean I should do the following:
```bash
cp /path/to/edge-login-ui-web/packages/edge-login-ui-web/assets/* /path/to/augur/src/assets
```
2. What do you mean by "it should work along with the bridge files"?  What is "it"?  Does that mean I need to also copy `packages/edge-login-ui-web/lib` somewhere?  If so, where?
3. Why is any of this necessary?

swansontec:
I re-worked the pull request with some changes:

1. The iframe contents are loaded from the Edge servers. I still think your installable desktop app should come with its own copy, but hopefully this solves the crashes you were experiencing.
2. The private keys are kept entirely within the iframe. The Edge login system signs all transactions.
3. The Edge login buttons now say "Loading..." while the frame is loading.
4. If as user has logged in with Edge before, we auto-open the Edge login window without requiring an extra button click.

Hopefully this all works seamlessly for you.

tinybike:
@swansontec when I try to use the "get rep" button after signing in, I get an "edge is signing" notification in my console, and a pending faucet notification inside the augur UI that has a timestamp of "11 hours ago", but the transaction never goes through, and I'm never presented with anything to sign / scan.  Am I doing something wrong?

tinybike:
Correction: the transactions actually did go through.  IIRC I'm not supposed to be prompted to scan anything in this version of the app, right?

swansontec:
Right, this version doesn't prompt the user for anything when signing.

tinybike:
@swansontec orders aren't registering correctly as belonging to the current user.  I think it's because `loginAccount.address` is supposed to be all lower case (`loginAccount.displayAddress` is mixed-case) -- it's a bit of technical debt we're going to pay off, but haven't just yet :/

swansontec:
Ok, I can fix the address thing pretty easily. Just `displayAddress.toLowerCase()` or whatever when I set up the account object.

tinybike:
Right, it's a simple fix -- just make sure the `address` is lower case, and `displayAddress` is mixed-case.

tinybike:
@swansontec I just tried to create an account through the Edge popup (inside Augur) and got this error:
```
create-ethrpc-configuration.js?5eeb:6 Error: unexpected EOF
    at errorHandler (blockchain-message-handler.js?2de1:45)
    at eval (blockchain-message-handler.js?2de1:54)
    at store.dispatch (index.js?3ae5:10)
    at Object.eval (reset-state.js?d2e8:45)
    at WebSocket.webSocketClient.onmessage (ws-transport.js?09f2:30)
```
After this, the UI lost its connection to my (local) Rinkeby Ethereum node, lost all its subscriptions, etc.

tinybike:
The account is created successfully -- I'm able to login to the UI using the account + PIN -- this is just something that happens right after the account creation.

tinybike:
Also @swansontec let's make it so that the Edge modal doesn't automatically pop up when you navigate to the accounts page.  (Popping up when you click on "connect an edge account" is fine, but popping up on navigating to the account page is a little awkward from a UX point of view.)

paullinator:
Hey @tinybike, is there a public node we can connect to for testing? I'd like to setup our QA team to be able to thoroughly hammer on this integration and make sure it's solid. That's the one hiccup for doing this since the app stopped connecting to public nodes?

I really want this to be solid for your release.

Thanks

tinybike:
Yes, there is!  We maintain public Rinkeby Ethereum nodes (geth) at `https://rinkeby.augur.net/ethereum-http` and `wss://rinkeby.augur.net/ethereum-ws` and public Rinkeby [Augur nodes](https://github.com/AugurProject/augur-node) at `wss://rinkeby.augur.nodes.augur.net`.

paullinator:
Ah sweet! Thanks. I'll get them on this.

tinybike:
@swansontec is Edge aware of what chain it is interacting with?  I'm testing things on Rinkeby and they seem to be working.  Should it work as-is on a private chain as well?

paullinator:
Signatures aren’t any different between networks right? Since the Augur code is the one interacting with the network, the Edge code cant actually tell the difference and is agnostic.

tinybike:
@swansontec I fixed up a couple things and submitted a PR (https://github.com/Airbitz/augur/pull/1) to the Airbitz/augur repo.  Is there a better workflow for submitting changes to a fork than this...?  Either way, I'm going to add you as a collaborator to the AugurProject/augur repo to avoid this awkwardness in future PRs.

swansontec:
I also added you as a collaborator to Airbitz/augur. To answer your earlier question, yes, the Edge login system is aware of the Ethereum network, thanks to the `data.chainId` that gets sent across with the transaction.

swansontec:
Rebased to clean up merge conflicts.

tinybike:
Looks good, @swansontec!  Very nice work getting this together.

PR #1374, opened 2018-04-06
Payout numerator stringify

pgebheim:


test/forking/actions/estimate-submit-migrate-rep-test.js
```diff
@@ -48,7 +48,7 @@ describe('modules/forking/actions/estimate-submit-migrate-rep.js', () => {
                 assert.deepEqual(args.tx, { to: '0xREP_TOKEN', estimateGas: true, gas: 600000 })
                 assert.equal(args.meta, 'META')
                 assert.equal(args._invalid, false)
-                assert.deepEqual(args._payoutNumerators, [0, 10000])
+                assert.deepEqual(args._payoutNumerators, ['0', '10000'])
```
nuevoalex: Will this actually work? If '10000' is passed as a string I think it will be evaluated as a hex string at the augur.js api level and so will be the wrong value

PR #1379, opened 2018-04-09
doInitialReport transaction failing

tinybike:
[Clubhouse Story](https://app.clubhouse.io/augur/story/8815)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


src/modules/reporting/actions/load-reporting-window-bounds.js
```diff
@@ -13,9 +13,9 @@ export const loadReportingWindowBounds = (callback = logError) => (dispatch, get
       if (err) return callback(err)
 
       dispatch(updateReportingWindowStats({
-        startTime: result.startTime,
-        endTime: result.endTime,
-        stake: result.totalStake || 0,
+        startTime: (result || {}).startTime,
```
justinbarry: Could this be rewritten: `(err, result={}) => {`?

tinybike: I don't think so, because `result` is an explicit `null` instead of not being passed at all.

justinbarry: Good call:
image

sharkcrayon:
This is working for me.

@justinbarry Feel free to merge when you think it's ready.

tinybike:
Not sure how this was working yesterday, but estimateSubmitInitialReport was being called _before_ the report form was filled out.  Fixed this so the estimate should work now!

PR #1408, opened 2018-04-15
Add `docker-compose pull` script to package.json

justinbarry:
[Clubhouse Story](https://app.clubhouse.io/augur/story/8971)


bthaile:
It should pull the image if not already pulled, right?

justinbarry:
Yes.

bthaile:
Why do we need this docker:spin-pull command, I'm a bit confused.

justinbarry:
@epheph Is the augur-node image tag stable?

justinbarry:
@bthaile :dev moves as augur-node is updated. We need this to pull newer images in.

bthaile:
I see what you are talking about. Also, I've seen the docker:spin-up fail because the images were taking too long to download. We need to build the pull into the spin-up and not have separate commands.

bthaile:
@justinbarry What do you think about my changes?

PR #1410, opened 2018-04-15
display the remaining REP on dispute form

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/8848)

Fixed displaying the correct value in dispute form. Simple change to display remaining REP instead of full bond amount.

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [x] Test/lint pass


sharkcrayon:
I think it’s best to display the total bond amount to the right of the
progress bars if that’s what this is referring to. Does someone think
differently?

bthaile:
I think differently, The Remaining Stake is the needed stake on the outcome to fulfil the current round bond. I think that is the number the user wants to see. If I understand, the total bond amount is the accumulative bond amount that spans over all dispute rounds.

sharkcrayon:
The total bond amount is particular to the round that the market is in. The
bond on each outcome is adjusted in each new dispute round the market
enters.

I view the total bond amount as the more useful piece of information to the
right of the progress bars. This is because we have the progress bars to
show percent progress: if the user sees that the total bond amount is 10
REP and the progress bar is halfway, then this signifies that there is
about 5 REP needed to fill the bond.

It is a bit ambiguous though.

sharkcrayon:
The progress bars denote “we are this far to reaching this goal amount”.

bthaile:
I see where you are coming from. Have to look at the calculations and what comes in from augur-node. My understanding is the total dispute bond is outcome specific. 

I was approaching it per dispute round not across dispute rounds. I'll see what exists.

bthaile:
We have an issue then, the progress starts over each round so I'm not calculating total percentage only per round. That is solvable pretty easily. I'll fix it.

sharkcrayon:
We do want it per round.

Each outcome has a dispute bond total that is particular to the dispute
round it is in. The progress bar is meant to show the progress made toward
filling that dispute bond during the current round.

Does that make sense? I can’t tell if we’re on the same page here, so sorry
if I’m being redundant.

bthaile:
The dispute bond grows as rounds get dispute bonds filled and a new winner is set.

Here is a short example, using data from get-dispute-info data from augur-node.
categorical market with 4 outcomes: DR picks an outcome 0 therefore putting 349680582682291667 atto rep on outcome. 
so the next amount is double with all other outcomes having 0 atto rep.
Round 1: 699361165364583334 is needed, all other outcomes have no progress towards completion. 
Let's say 0.5 REP is placed on outcome 1
so outcome 1 would have progress bar filled roughly 70%
Let's say invalid wins the round with 699361165364583334

Round 2 invalid is winner with 699361165364583334
outcome 0 has 349680582682291667
outcome 1 has 500000000000000000
New outcome with 0 stake needs (total existing dispute stake * 2) 2098083496093750002
outcome 0 needs (total existing dispute stake - it's already staked * 2) 1049041748046875001
outcome 1 needs (total existing dispute stake - it's already staked * 2) 199361165364583334

augur-node does all the calculations of outcome current bond size, account staked ...

My approach was to show Remaining REP needed to fill dispute round bond and the progression signified progress towards filling the current rounds bond. When they put in the amount of REP needed to fulfil the bond the number to the right goes to 0 and the progress bar fills. 

Side Note:
I think the users will want to know how much REP is needed for their outcome to win. I think that number needs to be displayed,  the progress bar is too ambiguous or a mouse over it to get the number is annoying. 

I can your proposal with showing the total possible amount needed to the right and have the progress bar show already staked REP (current user and other users). I'd propose see the progress bar height widened so we can display remaining REP needed for the dispute round. like we do with dispute window days remaining.

I think we have all the data we need to visualize the dispute progress bars. Currently the UI is caught between to different approaches. I'm up for whatever you decide. We can jump on a call anytime.

bthaile:
Maybe we can have the reporting dispute market card be overall outcome standings. The reporting dispute form can be current dispute round standings.

bthaile:
Updated this PR for reporting market card to show overall outcome dispute progress on market.

bthaile:
screen shot 2018-04-16 at 4 40 07 pm

changed all displays of dispute to be Current oriented, both dispute market card and dispute form.
added remaining REP in the dispute form, to look like dispute market card

bthaile:
screen shot 2018-04-16 at 9 26 49 pm


Made a style change to make the progress bar look more homogenous. Also see ch task for testing steps. I think we are good now, augur-node should be updated so that account stake current is correct. @sharkcrayon

bthaile:
ok, I got them backwards then. I put account staked there and the mouse over was overall staked rep.

sharkcrayon:
We're going to address what numbers are shown to the right of the progress bars in a different story. Merging this in.

PR #1413, opened 2018-04-17
johndanz/ch9062/when-canceling-an-open-order-confirmation

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9062/when-canceling-an-open-order-confirmation-message-should-include-average-price-os-the-order)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/market/components/market-positions-list--order/market-positions-list--order.jsx
+++ b/src/modules/market/components/market-positions-list--order/market-positions-list--order.jsx
@@ -14,8 +14,9 @@ export default class Order extends Component {
   static propTypes = {
     isExtendedDisplay: PropTypes.bool,
     isMobile: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/modules/market/components/market-positions-list--order/market-positions-list--order.jsx
```diff
@@ -14,8 +14,9 @@ export default class Order extends Component {
   static propTypes = {
     isExtendedDisplay: PropTypes.bool,
     isMobile: PropTypes.bool,
+    name: PropTypes.string,
```
adrake33: Maybe this should be called something like `outcomeName` to make it a little clearer what it represents?

PR #1422, opened 2018-04-18
Johndanz/ch6174/populate pending orders so that user can

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/6174/populate-pending-orders-so-that-user-can-t-close-position-when-pending-order)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/market/components/market-positions-list--position/market-positions-list--position.jsx
+++ b/src/modules/market/components/market-positions-list--position/market-positions-list--position.jsx
@@ -13,7 +13,7 @@ import Styles from 'modules/market/components/market-positions-list--position/ma
 
 export default class Position extends Component {
   static propTypes = {
-    name: PropTypes.string.isRequired,
+    outcomeName: PropTypes.string.isRequired,
     position: PropTypes.object.isRequired,
     openOrders: PropTypes.array.isRequired,
     isExtendedDisplay: PropTypes.bool.isRequired,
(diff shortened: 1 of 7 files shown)


src/modules/market/components/market-positions-list--position/market-positions-list--position.jsx
```diff
@@ -84,24 +87,24 @@ export default class Position extends Component {
         className={!isMobile ? Styles.Position : Styles.PortMobile}
       >
         <li>
-          { getValue(this.props, 'name') }
+          { outcomeName }
           { openOrders && openOrders.length > 0 && openOrders.map((order, i) => (
             <div key={i} className={Styles.Position__pending}>
-              <span className={Styles['Position__pending-title']}>Pending</span>
-              <span className={Styles['Position__pending-message']}>Bought { getValue(order, 'order.qtyShares.formatted') } at { getValue(order, 'order.purchasePrice.formatted') } ETH</span>
+              <span className={Styles['Position__pending-title']}>{order.pending ? `Pending` : `Open`}</span>
```
sharkcrayon: This is an interesting distinction. I think the purpose here is to show orders that have been filled but where that filled tx is still pending. If the order is not filled (still open) it will show in the Open Orders list, and if the order is about to become part of a position (because it has just been filled, but not confirmed) then it would show here.

Take a look at this and see what you think the original intention was:

screen shot 2018-04-18 at 5 56 39 pm

I don't think we'd want to show open orders in this location because then they'd be listed twice, once under Open Orders and once under My Positions under the related position.

PR #1427, opened 2018-04-18
update to fix using wrong button when market is awaiting next window …

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9156)

Fixed dangling market when last market is reported on. Also make sure button is correct for markets in awaiting next fee window.

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


bthaile:
Updated made to use logError, good call.

PR #1431, opened 2018-04-19
Fixing error caused by minPrice/maxPrice being null

adrake33:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9238/entering-number-of-shares-for-an-order-throws-error-in-scalar-markets)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


adrake33:
Ah, good point, @bthaile. I've removed that check from `generate-trade.js`.

PR #1448, opened 2018-04-24
johndanz/ch9289/can-t-trade-on-categorical-market

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9289/can-t-trade-on-categorical-market)

Removed the validation check, this is the fix suggested by Jack in the standup for now. There are other issues i feel need to be discussed but this narrow change has been made here.

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


JohnDanz:
Alright, Not only have we removed validation checks for multiples of tickSize on order price, but I've updated some math to use bignumbers to get around a depth chart rendering issue that kept cropping up. I've also changed how we decide what the estimated Cost value is going to be in updateTradeInProgress. After SimulatedTrade is run, we will run calculateTradeCost and then compare the value out of that to the tokensDepleted value sent back from simulateTrade. If the calculatedTradeCost is higher, we set the estimate to that. This way when augur.js gets estimatedTradeCost it has the proper value. Augur.js uses the estimated cost if it's not null, otherwise it defaults to the calculatedTradeCost.

PR #1456, opened 2018-04-26
Change to uPort's keypair account type

tinybike:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9345)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


src/modules/auth/actions/login-with-uport.js
```diff
@@ -1,18 +1,16 @@
...
-    dispatch(loadAccountData({
-      ...account,
-      meta: {
-        address: account.address,
-        signer: txObj => uPortSigner(txObj, dispatch),
-        accountType: augur.rpc.constants.ACCOUNT_TYPES.U_PORT,
-      },
-    }))
-  }
+export const loginWithUport = account => (dispatch) => {
```
adrake33: Why was this changed from being an `export default`? I thought we were trying to use `export default` whenever possible.

tinybike: Just for consistency, generally export const is used else where.

PR #1460, opened 2018-04-26
johndanz/ch9455/bad-position-shares-math

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/9455/bad-position-shares-math)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/my-positions/selectors/my-positions-summary.js
+++ b/src/modules/my-positions/selectors/my-positions-summary.js
@@ -10,17 +10,24 @@ import { ZERO } from 'modules/trade/constants/numbers'
 // import { augur } from 'services/augurjs'
 import { formatEther, formatShares, formatNumber } from 'utils/format-number'
 
-export const generateOutcomePositionSummary = memoize((adjustedPosition) => {
+export const generateOutcomePositionSummary = memoize((adjustedPosition, openOrders) => {
   if (!adjustedPosition) {
     return null
   }
+  const simplifiedOpenOrders = openOrders.reduce((accum, item) => {
+    accum.push({ unmatchedShares: item.unmatchedShares.fullPrecision })
+    return accum
+  }, [])
   const outcomePositions = Array.isArray(adjustedPosition) ? adjustedPosition.length : 1
+  
+  const qtySharesOpen = accumulate(simplifiedOpenOrders, 'unmatchedShares')
   const qtyShares = accumulate(adjustedPosition, 'numSharesAdjusted')
+  const adjustedQtyShares = qtyShares.plus(qtySharesOpen)
   const realized = accumulate(adjustedPosition, 'realizedProfitLoss')
   const unrealized = accumulate(adjustedPosition, 'unrealizedProfitLoss')
   // todo: check if this calculation is correct for UI
   const averagePrice = accumulate(adjustedPosition, 'averagePrice')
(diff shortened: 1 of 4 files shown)


src/modules/my-positions/selectors/my-positions-summary.js
```diff
@@ -6,21 +6,34 @@ import store from 'src/store'
...
 
 // import { augur } from 'services/augurjs'
 import { formatEther, formatShares, formatNumber } from 'utils/format-number'
 
-export const generateOutcomePositionSummary = memoize((adjustedPosition) => {
+export const generateOutcomePositionSummary = memoize((adjustedPosition, openOrders) => {
   if (!adjustedPosition) {
     return null
   }
+  const simplfiedOrders = openOrders.reduce((accum, item) => {
```
tinybike: Fix spelling to `simplifiedOrders`.

src/modules/my-positions/selectors/my-positions-summary.js
```diff
@@ -6,21 +6,34 @@ import store from 'src/store'
...
+    })
+    return accum
+  }, { [BUY]: [], [SELL]: [] })
+
   const outcomePositions = Array.isArray(adjustedPosition) ? adjustedPosition.length : 1
+  const qtySellSharesOpen = accumulate(simplfiedOrders[SELL], 'unmatchedShares')
+  const qtyBuySharesOpen = accumulate(simplfiedOrders[BUY], 'unmatchedShares')
   const qtyShares = accumulate(adjustedPosition, 'numSharesAdjusted')
+
+  const adjustedQtyShares = qtyShares.isPositive() ? qtyShares.minus(qtySellSharesOpen) : qtyShares.plus(qtyBuySharesOpen)
```
tinybike: Two comments here:

1. I think this adjustment should not be necessary (if I understand how the share balance lookups work correctly).  If the share balances are not being adjusted appropriately when orders are created, that's a bug and should be addressed directly, instead of papering over it by doing an adjustment here.
2. Apart from being unnecessary, I don't think this adjustment is being done correctly here.  The only thing that should affect the user's position is open orders that have shares escrowed.  But again, this adjustment shouldn't be necessary at all, unless I'm misunderstanding something.

tinybike:
@JohnDanz I left a comment about the position adjustments in this PR and why I don't think they're necessary (or correct).  Another meta-comment is that the position adjustment stuff seems like it would belong in a different PR, as it appears to be out of the scope of [ch9455](https://app.clubhouse.io/augur/story/9455/bad-position-shares-math).  Or, am I misunderstanding?

bthaile:
https://app.clubhouse.io/augur/story/9455/bad-position-shares-math

added some screenshots of the 9.99... display issue. John pointed it out, but there is a difference how the Order Book shows the 10 share order and how the Outcome List shows the Order.
I think we need to be consistent.

JohnDanz:
@tinybike I actually figured it went here since it wasn't updating correctly so it was "bad shares math" But i agree with your meta comment that it probably should be cleaned up elsewhere then in the selector so I'm going to pull out the changes I made to try and account for orders and create a new task to investigate why it appears to not be updating correctly. 

So this next commit should have just the changes to use BigNumber Math in the accumulator. 

@bthaile you're comment is OOS for the same reason as above, we should make a seperate issue to investigate that, I'll create a clubhouse issue for that.

PR #1477, opened 2018-05-01
Johndanz/ch6174/populate pending orders so that user can

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/6174/populate-pending-orders-so-that-user-can-t-close-position-when-pending-order)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


JohnDanz:
Changes include:

Positions should now be closable
updated positions text to say "at market price" not a value
"pending cancellation" text added to open orders only after the user has signed and sent their cancel order transaction.

PR #1553, opened 2018-05-11
Justinbarry/ch10078/market liquidity depth chart jumps on hover

justinbarry:
[Clubhouse Story](https://app.clubhouse.io/augur/story/10078)


justinbarry:
@bthaile This makes sense. I'll merge it down.

PR #1584, opened 2018-05-14
Fix candlestick x-axis

justinbarry:


JohnDanz:
@justinbarry this looks like a variable name change to me, how does this effect the x-axis?

PR #1632, opened 2018-05-17
Escape hatch cancel orders

nuevoalex:
[Clubhouse Story](https://app.clubhouse.io/augur/story/10734)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/escape-hatch/actions/withdraw-funds-in-emergency.js
+++ b/src/modules/escape-hatch/actions/withdraw-funds-in-emergency.js
@@ -6,12 +6,14 @@ import { updateMarketRepBalance, updateMarketEthBalance, updateMarketFrozenShare
 import { updateParticipationTokenBalance } from 'modules/my-participation-tokens/actions/update-participation-tokens'
 import { updateInitialReporterRepBalance } from 'modules/my-initial-reporters/actions/update-initial-reporters'
 import { updateDisputeCrowdsourcersBalance } from 'modules/my-dispute-crowdsourcer-tokens/actions/update-dispute-crowdsourcer-tokens'
+import { updateOrderClearEscrowed } from 'modules/escape-hatch/actions/update-all-orders'
+import { doUpdateShareFrozenValue } from 'modules/escape-hatch/actions/load-emergency-withdrawal-assets'
 
 export default function (ownedMarkets, marketsWithShares, callback = logError) {
   return (dispatch, getState) => {
 
     const {
-      participationTokens, initialReporters, disputeCrowdsourcerTokens = {}, loginAccount,
+      participationTokens, initialReporters, disputeCrowdsourcerTokens, allOrders = {}, loginAccount,
     } = getState()
 
     each(ownedMarkets, (market) => {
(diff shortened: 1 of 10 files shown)


nuevoalex:
To test this you'll need to either use this branch of AN or wait till it gets merged:

https://github.com/AugurProject/augur-node/pull/361

src/modules/escape-hatch/actions/withdraw-funds-in-emergency.js
```diff
@@ -85,5 +87,33 @@ export default function (ownedMarkets, marketsWithShares, callback = logError) {
...
+          dispatch(updateOrderClearEscrowed(orderId))
+          if (orderHasSharesEscrowed) {
+            dispatch(doUpdateShareFrozenValue(order.marketId, dispatch, () => {
+              augur.api.TradingEscapeHatch.claimSharesInUpdate({
+                meta: loginAccount.meta,
+                _market: order.marketId,
+                onSent: noop,
+                onSuccess: (res) => {
+                  console.log('TradingEscapeHatch.claimSharesInUpdate', res)
+                  dispatch(updateMarketFrozenSharesValue(order.marketId, 0))
```
bthaile: probably don't need this console log.

src/modules/escape-hatch/selectors/escape-hatch-data.js
```diff
@@ -71,7 +76,14 @@ export const getEscapeHatchData = createSelector(
...
 
-    data.fundsAvailableForWithdrawal = data.rep + data.eth > 0
+    Object.keys(orders).forEach((orderId) => {
+      const order = orders[orderId]
+      data.eth = data.eth.plus(createBigNumber(speedomatic.fix(order.tokensEscrowed)))
+      data.shares = data.shares.plus(createBigNumber(speedomatic.fix(order.sharesEscrowed)))
+      data.gas = data.gas.plus(CANCEL_ORDER_GAS_ESTIMATE)
+    })
+
+    data.fundsAvailableForWithdrawal = data.rep + data.eth + data.shares > 0
```
bthaile: is '+' same as .plus with BigNumbers?

`fundsAvailableForWithdrawal` seems to be used as a number but is a bool. 

referencing src/modules/escape-hatch/components/escape-hatch.jsx
`  const fundsAvailableForWithdrawal = nextProps.escapeHatchData.fundsAvailableForWithdrawal > 0`

How about using it as a bool or use it as a number. If use as number use .plus(....).plus(...).toNumber()

bthaile:
Created an order for 100 shares at .1, the `get-market-balance` script says I have 100 shares.

After pulling escape hatch, I expected to see Shares of 100.

Withdrawing funds gives the correct amount of ETH. Might be I'm missing the correct trading setup.

screen shot 2018-05-17 at 9 45 40 am


screen shot 2018-05-17 at 9 46 03 am

src/modules/escape-hatch/selectors/escape-hatch-data.js
nuevoalex: ah yea good point, will do

bthaile:
`const fundsAvailableForWithdrawal = nextProps.escapeHatchData.fundsAvailableForWithdrawal > 0`
does this need to be fundsAvailableForWithdrawal.gt(ZERO) // ZERO using constant

`import { ZERO } from 'modules/trade/constants/numbers'`
just for convention.

PR #1644, opened 2018-05-17
fix for weird behavior when switching accounts

nuevoalex:
[Clubhouse Story](https://app.clubhouse.io/augur/story/10968)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


sharkcrayon:
@nuevoalex Yeah, can you make it not round up in all cases? 0.00034 and below should round down to 0.0003, while 0.00035 and up should round to 0.0004.

Also, I'm seeing the same balance before and after pulling the escape hatch. Retesting one more time to make sure.

PR #1672, opened 2018-05-19
achapman/ch11165/trades-which-involve-the-sale-of-shares-charge

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11165/trades-which-involve-the-sale-of-shares-charge-too-much-eth-and-refund-excess)

Goes With these AugurJS changes: https://github.com/AugurProject/augur.js/pull/391

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/trade/actions/place-trade.js
+++ b/src/modules/trade/actions/place-trade.js
@@ -17,12 +17,16 @@ export const placeTrade = (marketId, outcomeId, tradeInProgress, doNotCreateOrde
     return dispatch(clearTradeInProgress(marketId))
   }
   const bnAllowance = createBigNumber(loginAccount.allowance)
+  const sharesDepleted = createBigNumber(tradeInProgress.sharesDepleted, 10)
+  const otherSharesDepleted = createBigNumber(tradeInProgress.otherSharesDepleted, 10)
+  const sharesProvided = sharesDepleted.eq(0) ? otherSharesDepleted.toString() : sharesDepleted.toString()
   // try and make sure that we actually have an updated allowance.
   if (bnAllowance.lte(0)) dispatch(checkAccountAllowance())
   const placeTradeParams = {
     meta: loginAccount.meta,
     amount: tradeInProgress.numShares,
     limitPrice: tradeInProgress.limitPrice,
+    sharesProvided,
     minPrice: market.minPrice,
     maxPrice: market.maxPrice,
     numTicks: market.numTicks,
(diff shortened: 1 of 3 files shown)


src/modules/trade/actions/place-trade.js
```diff
@@ -17,12 +17,16 @@ export const placeTrade = (marketId, outcomeId, tradeInProgress, doNotCreateOrde
     return dispatch(clearTradeInProgress(marketId))
   }
   const bnAllowance = createBigNumber(loginAccount.allowance)
+  const sharesDepleted = createBigNumber(tradeInProgress.sharesDepleted, 10)
+  const otherSharesDepleted = createBigNumber(tradeInProgress.otherSharesDepleted, 10)
+  const sharesProvided = sharesDepleted.eq(0) ? otherSharesDepleted.toString() : sharesDepleted.toString()
```
tinybike: Does `toString` always do what we want here?  Normally we used `toFixed` for base-10 string conversion.  IIRC `toString` can use scientific notation for large values and other weirdness that we don't want here.

src/modules/trade/actions/place-trade.js
```diff
@@ -17,12 +18,16 @@ export const placeTrade = (marketId, outcomeId, tradeInProgress, doNotCreateOrde
     return dispatch(clearTradeInProgress(marketId))
   }
   const bnAllowance = createBigNumber(loginAccount.allowance)
-  // try and make sure that we actually have an updated allowance.
-  if (bnAllowance.lte(0)) dispatch(checkAccountAllowance())
+  const sharesDepleted = createBigNumber(tradeInProgress.sharesDepleted, 10)
+  const otherSharesDepleted = createBigNumber(tradeInProgress.otherSharesDepleted, 10)
+  const sharesProvided = sharesDepleted.eq(ZERO) ? otherSharesDepleted.toFixed() : sharesDepleted.toFixed()
+  // make sure that we actually have an updated allowance.
+  if (bnAllowance.lte(ZERO)) dispatch(checkAccountAllowance())
```
justinbarry: Is this async? If so, should we put the rest of this in a callback?

JohnDanz: separate issue, planning to look at that today, it's the approval thing. That will probably go away or change slightly. That dispatch is intended to make sure we actually pulled the latest allowance but you are correct it shouldn't be async and that's probably what will change. feel free to ignore here though as it'll be in it's own PR

PR #1680, opened 2018-05-20
Fix market depth chart text and nearest order calculations

justinbarry:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11170)

## Submitter checklist


bthaile:
Found a empty item in the order book where there's no price, share amount or depth, hovering over it highlights the depth chart.


screen shot 2018-05-20 at 6 18 05 pm

bthaile:
that makes sense, good call

PR #1702, opened 2018-05-22
don't set designated stake from market creation costs if it's an open…

bthaile:
… reporting initial report, use market no-show bond amount

[Clubhouse Story](https://app.clubhouse.io/augur/story/11465)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/reporting/components/reporting-report/reporting-report.jsx
+++ b/src/modules/reporting/components/reporting-report/reporting-report.jsx
@@ -50,7 +50,7 @@ export default class ReportingReport extends Component {
         selectedOutcome: null,
       },
       reporterGasCost: null,
-      designatedReportNoShowReputationBond: 0,
+      designatedReportNoShowReputationBond: formatEtherEstimate(props.market.designatedReportStake),
       gasEstimate: '0',
     }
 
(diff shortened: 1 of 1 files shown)


sharkcrayon:
This isn't working properly for me. Steps to test:

Increase the no-show bond amount paid by the market creator:

1. Start the environment up fresh
2. Move time forward by 1 day. Three markets enter DR state.
3. Move time forward by 3 days. All (three) markets in DR are moved to OR.
4. With a user that is _not_ the DR, report on all markets in OR.
5. Move time forward by 7 days so next fee window starts and the markets move into Dispute.
6. Move time forward by 7 days so the next fee window starts and the markets resolve.
7. Claim your reporting fees so the markets finalize.
8. Walk through the create market form. Set your end date to January 1st. Look at fee amounts on Review page. It should be 0.6994 REP. Create your market (and use a name you'll remember).

Now test that the no-show bond paid at market creation is shown to an OR on the review screen when they report:

1. Move time forward to January 1st so your market moves into DR, along with other markets that were created at initial deploy.
2. Move time forward 3 days so the markets move into OR.
3. View the Reporting Form review screen for a market that was created with the initial deploy. The text should say the OR will receive at least 0.3497 REP.
4. [THIS STEP FAILS] View the Reporting Form review screen for the market you created with the higher no-show bond amount. The text should say the OR will receive at least 0.6994 REP.

PR #1712, opened 2018-05-25
UI will send failing tx for create scalar market If numTicks is odd

adrake33:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11536/ui-will-send-failing-tx-for-create-scalar-market-if-numticks-is-odd)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
+++ b/src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
@@ -153,6 +153,7 @@ export default class CreateMarketOutcome extends Component {
     // reset errors
     updatedMarket.validations[currentStep].scalarBigNum = true
     updatedMarket.validations[currentStep].scalarSmallNum = true
+    updatedMarket.validations[currentStep].tickSize = true
 
     switch (true) {
       case scalarSmallNum === '':
@@ -207,6 +208,11 @@ export default class CreateMarketOutcome extends Component {
           updatedMarket.validations[currentStep].tickSize = true
       }
       updatedMarket.tickSize = value
+    }
+
+    // Always check if (maxPrice - minPrice) / precision is an even number
+    if (((scalarBigNum - scalarSmallNum) / numTicksBigNum) % 2 !== 0) {
+      updatedMarket.validations[currentStep].tickSize =`Total ticks must be even.`
     }
 
     updatedMarket.isValid = isValid(currentStep)
--- a/src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
+++ b/src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
@@ -214,7 +214,7 @@ export default class CreateMarketOutcome extends Component {
     if (BigNumber.isBigNumber(scalarBigNum) && BigNumber.isBigNumber(scalarSmallNum) && BigNumber.isBigNumber(numTicksBigNum)) {
       // Always check if (maxPrice - minPrice) / precision is an even number
       if ((scalarBigNum.minus(scalarSmallNum).div(numTicksBigNum)).mod(2).toString() !== '0') {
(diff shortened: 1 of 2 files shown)


src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
```diff
@@ -209,6 +210,11 @@ export default class CreateMarketOutcome extends Component {
       updatedMarket.tickSize = value
     }
 
+    // Always check if (maxPrice - minPrice) / precision is an even number
+    if (((scalarBigNum - scalarSmallNum) / numTicksBigNum) % 2 !== 0) {
```
epheph: This isn't safe for BigNumbers

adrake33: I was originally using `(scalarBigNum.sub(scalarSmallNum).div(numTicksBigNum)).mod(2) !== 0` in the switch statement above, which wasn't working. (I realize now I should have been using `minus` instead.) I had temporarily switched to using regular math operators just to see if moving it below the switch statement would work the way I wanted, but forgot to switch it back to using normal big number operators. Will update.

adrake33: Actually doing `console.log(typeof scalarBigNum)` outputs `string` here, so I'll need to make sure all of these values are BigNumbers at this point in the code.

adrake33: Ah, the issue is that `scalarBigNum` is the empty string when the page first loads, so it's not getting set to a BigNumber up above. I'll add a check to make sure all of the numbers are actually BigNumbers.

adrake33:
@bthaile In the example you gave, the user should be allowed to submit the form. With this new change, (maxPrice - minPrice) / precision must be an even number, and in this case (5 - 0) / 1.25 = 4.

bthaile:
I got on the newest of now contracts and it all looks good.

adrake33:
@sharkcrayon Is the error message "Increase range or precision." ok in this case? It's not great, but it's the only thing that fits under the textbox. Alternatively, we could make it a link that goes to the docs.

src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
```diff
@@ -209,6 +210,14 @@ export default class CreateMarketOutcome extends Component {
       updatedMarket.tickSize = value
     }
 
+    // Make sure scalarBigNum, scalarSmallNum, & numTicksBigNum are all BigNumbers
+    if (BigNumber.isBigNumber(scalarBigNum) && BigNumber.isBigNumber(scalarSmallNum) && BigNumber.isBigNumber(numTicksBigNum)) {
```
JohnDanz: why are we checking that these are big numbers before deciding if they have a proper numTickSize?

src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
```diff
@@ -209,6 +210,14 @@ export default class CreateMarketOutcome extends Component {
...
+      // Always check if (maxPrice - minPrice) / precision is an even number
+      if ((scalarBigNum.minus(scalarSmallNum).div(numTicksBigNum)).mod(2).toString() !== '0') {
```
JohnDanz: This looks like validation that should/could be performed in the switch, is there a reason we do this separately from the above switch?

adrake33: I originally tried putting this check in the switch statement above, but the error message needs to get updated as the user is typing in the "Precision" textbox, as well as the "Min Value"/"Max Value" textboxes, since all of these values are used to determine the number of ticks.

PR #1717, opened 2018-05-26
found a few issues, one the reducer wasn't keying off of marketId so …

bthaile:
…close position status would fail. Found close position was updating when position status was undefined when it was null

[Clubhouse Story](https://app.clubhouse.io/augur/story/11552)

Also added an message to inform user a closing their position failed, before it was showing the confirmation message which is wrong.

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


bthaile:
Kicked off CI again see if it passes. fingers-crossed

PR #1727, opened 2018-05-28
update assets on block updates instead of spamming it on some blocks

nuevoalex:
Did essentially what we discussed. Works well.

[Clubhouse Story](https://app.clubhouse.io/augur/story/11696)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


tinybike:
This is a bad fix and should be reverted IMO.

tinybike:
It would be really good to fix this *without* calling updateAssets on every single block.  There's a lot of overhead associated with that that we've made a point of avoiding by not doing this.

tinybike:
(I'm assuming the overall impact of updateAssets on every block is greater than calling it multiple times if you happen to get a block with multiple updateAssets-calling log processors -- is that incorrect?)

nuevoalex:
On mainnet blocks come in every 15 seconds. Even on private chain they come in every 2 seconds or so. Why do you think this is incurring large overhead?

PR #1728, opened 2018-05-28
0 is a vaild value for scalar markets, use '-'for default formatted v…

bthaile:
…alue when midpoint is selected and there are no trades

[Clubhouse Story](https://app.clubhouse.io/augur/story/11624)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/market/selectors/market.js
+++ b/src/modules/market/selectors/market.js
@@ -236,12 +236,12 @@ export function assembleMarket(
           ...outcomeData,
           id: outcomeId,
           marketId,
-          lastPrice: formatEther(outcomeData.price || 0, { positiveSign: false }),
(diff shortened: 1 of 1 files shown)


src/modules/market/selectors/market.js
```diff
@@ -236,19 +236,21 @@ export function assembleMarket(
           ...outcomeData,
           id: outcomeId,
           marketId,
-          lastPrice: formatEther(outcomeData.price || 0, { positiveSign: false }),
+          lastPrice: outcomeData.price ? formatEther(outcomeData.price, { positiveSign: false }) : null,
```
JohnDanz: if outcomeData.price === 0 this will return null cuz 0 is false-y. 0 might actually be a legitimate price.

PR #1731, opened 2018-05-28
johndanz/ch11553/pl-scale-adjustment

JohnDanz:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11553/pl-scale-adjustment)

see ticket for reproduction instructions. this should fix all instances though as i erroneously was rounding up on the max value, which is why the scale was so off. (showing 1+ ETH top end even if our value was 0.00019, while making 0 the bottom end...)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


JohnDanz:
@bthaile so this is actually a separate issue which i made a ticket for with paul and discussed. We plan to get cracking on that today, however based on that screenshot above (first one) i do need to still make a small adjustment, apparently this issue isn't 100% fixed for all cases. will re-open the PR when it's ready

PR #1750, opened 2018-05-31
Johndanz/ch11649/trade page maxes out cpu when left open

justinbarry:
[Clubhouse Story](https://app.clubhouse.io/augur/story/ch11649)


sharkcrayon:
Clubhouse story: https://app.clubhouse.io/augur/story/11649/trade-page-maxes-out-cpu-when-left-open

(The link in the original comment was failing)

PR #1757, opened 2018-05-31
Check for matching network IDs if using global web3

tinybike:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11937)

Note: requires https://github.com/AugurProject/augur.js/pull/416 to be merged first

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


MicahZoltu:
Is `isMetaMask` still used?  Looks like the file was retained but all references were changed to global web3.

If it is used, what is it being used for?

tinybike:
Yes, it is used to detect if using MetaMask so the wording of the network mismatch modal can be specific to MetaMask.

PR #1770, opened 2018-06-02
Fix mist not logging in.

Smithgift:
With this change, mist appears to work out-of-the-box, without having to
manually unlock an account. Note that you'll need to use a full client for the moment or it won't work anyway.

_However,_ I could not get the tests to pass on my own, and I believe doing so would require changes perhaps even to `ethrpc`, which would be beyond my comfort zone. @MicahZoltu suggested leaving it up to you guys, so here it is.

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [x] Test/lint pass


MicahZoltu:
cc @tinybike since I believe you are most familiar with this code.

tinybike:
Hm, that's strange, it looks like the Travis build is trying to pick up a nonexistent version of ethrpc for some reason.  It works fine locally.

tinybike:
augur.js version may just need to be bumped...

src/modules/auth/actions/use-unlocked-account.js
```diff
@@ -1,14 +1,15 @@
...
+import isGlobalWeb3 from 'modules/auth/helpers/is-global-web3'
 import logError from 'utils/log-error'
 
 const { ACCOUNT_TYPES } = augur.rpc.constants
 
 // Use unlocked local account or MetaMask account
 export const useUnlockedAccount = (unlockedAddress, callback = logError) => (dispatch) => {
   if (unlockedAddress == null) return callback('no account address')
-  if (isMetaMask()) {
+  if (isGlobalWeb3() || isMetaMask()) {
```
tinybike: I think `if (isGlobalWeb3())` should work fine here; `isMetaMask()` actually does `isGlobalWeb3()` inside of it, so the `||` will never be hit!

tinybike:
@Smithgift can you try running `npm install --save-exact augur.js@dev` then pushing the updated package.json and lockfiles up to your branch?  I think that will fix the Travis error you're seeing.

Smithgift:
Blergh, I think yarn needs updating too. One moment. Apologies for all these extra commits.

(NINJA EDIT: And now yarn is down, too. :rage: )

In, re: `isGlobalWeb3() || isMetaMask()` these [tests](https://github.com/AugurProject/augur/blob/master/test/auth/actions/use-unlocked-account-test.js) failed until I changed it from what you suggested into this. I believe, on further thought, it's the test's fault, because they have no reference to a world with `isGlobalWeb3()`. For that matter, there's no account type of `LOCKED_BUT_FUNCTIONING_GLOBAL_WEB3` in ethrpc. Not sure what you would like to have happen there.

tinybike:
@Smithgift I see you've had your first run-in with our Mock All The Things unit tests.  I think the test failures were just because `isGlobalWeb3` was not stubbed out in the tests.

tinybike:
@Smithgift I submitted a PR to your fork (https://github.com/Smithgift/augur/pull/1) which removes the `|| isMetaMask()` check and adds `isGlobalWeb3` stubs for the unit tests.

Side note: is this really the expected workflow for editing PRs submitted from forks?  Very clunky IMO :/

Smithgift:
Thanks for fixing things! I confirmed that it still works on a private network mist.

In, re: the PR thing, I haven't a clue. This will be my third pull request ever--I mostly develop alone.

PR #1772, opened 2018-06-03
Tomhaile/ch11995/need to listen for market migrated log to

bthaile:
[Clubhouse Story](https://app.clubhouse.io/augur/story/11995)

## Submitter checklist
  - [x] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass


src/modules/events/actions/log-handlers.js
```diff
@@ -40,6 +40,15 @@ export const handleMarketCreatedLog = log => (dispatch, getState) => {
   }
 }
 
+export const handleMarketMigratedLog = log => (dispatch, getState) => {
+  if (log.removed) {
+    dispatch(loadMarketsInfo([log.market]))
+  } else {
+    dispatch(removeMarket(log.market))
```
nuevoalex: I think instead of only filtering based on the original universe it should filter on if the current universe is _either_ the original or new and then here it should add or remove based on that.

bthaile: ok, will make corrections

PR #1778, opened 2018-06-03
Add default boundDiff and set volume min to zero

justinbarry:
[Clubhouse Story](https://app.clubhouse.io/augur/story/12009)


bthaile:
@justinbarry made a comment on CH task. There might be a small issue with Month time ranges at the bottom of the candlestick chart that makes the data look bad, but it's probably something simple.

PR #1817, opened 2018-06-08
use publicCreateOrder for market liquidity

nuevoalex:
[Clubhouse Story](https://app.clubhouse.io/augur/story/12290)

## Submitter checklist
  - [ ] Steps to reproduce

## Reviewer Checklist
- [ ] Test/lint pass

### DIFF
--- a/src/modules/create-market/actions/submit-new-market.js
+++ b/src/modules/create-market/actions/submit-new-market.js
@@ -64,10 +64,11 @@ export function submitNewMarket(newMarket, history) {
                   },
                   onFailed: (err) => {
                     console.error('ERROR creating order in initial market liquidity: ', err)
+                    orderCB()
                   },
                 })
               }, (err) => {
-                if (err !== null) return console.error('ERROR: ', err)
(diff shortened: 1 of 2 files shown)


src/modules/create-market/actions/submit-new-market.js
```diff
@@ -39,22 +37,37 @@ export function submitNewMarket(newMarket, history) {
...
+                  _market: marketId,
+                  _outcome: outcomeId,
+                  _tradeGroupId: augur.trading.generateTradeGroupId(),
+                  onSent: noop,
+                  onSuccess: (res) => {
+                    orderCB()
+                  },
+                  onFailed: (err) => {
+                    console.error('ERROR creating order in initial market liquidity: ', err)
+                  },
```
bthaile: Is orderCD() needed if publicCreateOrder fails? seems like it would be in waiting state.

nuevoalex: Yea I think you're right!. Weirdly that doesn't appear to be how the old behavior worked

PR #1863, opened 2018-06-19
Set MetaMask as default account connect; remove uPort integration

tinybike:


adrake33:
@tinybike I believe you said in today's standup that these changes are for MVP, but this PR is for merging into master. Should I still go ahead and merge this to master?  (I know you also said we should just have a single branch, though, so maybe it doesn't really matter since they will ultimately become one branch.)

paullinator:
My I selfishly 😉 put in a vote to keep Edge as the default account mechanism. Thoughts are that any other option, including Metamask, would require the user to have Metamask installed which is limited to Chrome/Firefox and only to desktop browsers. Full size tablet browsers, Safari on MacOS, and MS Edge would be incompatible. Edge Login, being built into Augur, doesn't require any external app.

paullinator:
@tinybike Totally agree with the auto-popup. I'm fine having that disabled and I thought you pushed a PR that did exactly that. If not, I'll have @swansontec submit a PR. 

A year ago when we had the auto popup, it was only meant to happen if the user already had used Edge and had an existing account on their browser. It was never meant to always pop up. Sorry if that snuck in. We'll have any auto popup removed.

adrake33:
I tried changing the base for this PR to mvp, but there were numerous conflicts, so I just decided to create a new PR to replace this one: https://github.com/AugurProject/augur/pull/1868

PR #1866, opened 2018-06-19
Tomhaile/ch12643/transfer transactions are confusing

bthaile:
[CH Task](https://app.clubhouse.io/augur/story/12643/transfer-transactions-are-confusing)

make sure no Transfer with weird long number of shares occurs in transactions page.

trade with second user then make order using shares and have other user take that trade.
1. Second User take canned market order
2. Second User use shares to create order
3. First User take the new order

verify no transfers like in the screen shot of CH task.

Need augur-node PR merged in, it has been approved. [augur-node PR](https://github.com/AugurProject/augur-node/pull/458)


bthaile:
Need to investigate another case not covered by PR.

PR #1873, opened 2018-06-20
johndanz/ch12730/filled-orders-showing-up-in-order-book

JohnDanz:
https://app.clubhouse.io/augur/story/12730/filled-orders-showing-up-in-order-book


bthaile:
What about log-handler.js it doesn't pass in load only OPEN orders?

screen shot 2018-06-21 at 10 50 07 am

PR #1875, opened 2018-06-21
only sort alphabetically

phoebemirman:
https://app.clubhouse.io/augur/story/12731/outcome-order-sequence


bthaile:
does case matter in sorting? 

screen shot 2018-06-21 at 1 04 38 pm

PR #1887, opened 2018-06-22
Phoebemirman/ch12797/market liquidity validation is off

phoebemirman:
https://app.clubhouse.io/augur/story/12797/market-liquidity-validation-is-off

### DIFF
--- a/src/modules/create-market/components/create-market-form-liquidity/create-market-form-liquidity.jsx
+++ b/src/modules/create-market/components/create-market-form-liquidity/create-market-form-liquidity.jsx
@@ -322,11 +322,7 @@ export default class CreateMarketLiquidity extends Component {
     let isOrderValid
 
     // Validate Quantity
-    if (orderQuantity !== '' && orderPrice !== '' && orderPrice.times(orderQuantity).plus(newMarket.initialLiquidityEth).gt(createBigNumber(availableEth))) {
-      // Done this way so both inputs are in err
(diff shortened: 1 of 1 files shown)


src/modules/create-market/components/create-market-form-liquidity/create-market-form-liquidity.jsx
```diff
@@ -382,8 +378,15 @@ export default class CreateMarketLiquidity extends Component {
         singleOutcomeOrderBook: newMarket.orderBook[outcome] || {},
       }
       const action = augur.trading.simulateTrade(orderInfo)
-      if (createBigNumber(action.tokensDepleted, 10).lt(tickSize)) {
-        errors.price.push(`Est. Cost of trade must be at least ${tickSize}`)
+      if (createBigNumber(action.tokensDepleted, 10).lt(0.0001)) {
+        errors.price.push('Est. Cost of trade must be at least 0.0001')
```
bthaile: need to use tickSize, here is an example that exists in code.
`errors.price.push(`Price must be a multiple of ${tickSize}`)`

bthaile: Suggest removing this check, checking against tokensDepleted isn't needed.

tinybike:
Does this need to also be reviewed by @JohnDanz?

PR #1888, opened 2018-06-22
integrate AirSwap widget

grahamperich:
This PR implements the AirSwap widget, which allows users to swap ETH for REP atomically with zero fees. Users just pay the gas cost of the transaction. 

I built Augur locally for development, but I also tested the integration by performing a swap on the Ethereum main net. I accomplished this by doing a little DOM hacking with the Chrome inspector. You can see how I tested it in this short screencast:

https://streamable.com/c99ay

Please let me know if there are any adjustments you'd like on this PR. We are big fans of Augur, and very excited for the main net launch! AirSwap also has some [developer resources](https://github.com/airswap/developers) if you're interested in further collaboration down the road.

### DIFF
--- a/src/modules/account/components/account-deposit/account-deposit.jsx
+++ b/src/modules/account/components/account-deposit/account-deposit.jsx
@@ -9,6 +9,20 @@ import { Deposit as DepositIcon, Copy as CopyIcon } from 'modules/common/compone
 
 import Styles from 'modules/account/components/account-deposit/account-deposit.styles'
 
+function airSwapOnClick(e) {
+  e.preventDefault()
+  window.AirSwap.Trader.render({
+    mode: 'buy',
+    token: '0xe94327d07fc17907b4db788e5adf2ed424addff6',
+    onCancel() {
+      console.info('AirSwap trade cancelled')
+    },
+    onComplete(txid) {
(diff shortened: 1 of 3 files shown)


src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -48,6 +64,16 @@ export default class AccountDeposit extends Component {
           </button>
         </div>
       )
+      airSwapConverter = (
+        <div className={Styles.AccountDeposit__shapeShiftButton}>
+          <button
+            onClick={e => airSwapOnClick(e)}
+            value={'https://shapeshift.io/shifty.html?destination=' + address + '&output=REP'}
```
tinybike: Hm, is this intended to be a ShapeShift link...?

grahamperich: fixed, thank you.

JohnDanz:
Is there any way to test this on Rinkeby? I'd rather not have to convert my real ETH/REP to test...

grahamperich:
@JohnDanz I pushed a new commit which will point the widget to the airswap rinkeby sandbox when web3 is not set to main net. You can now test the functionality with testnet ETH.

PR #1892, opened 2018-06-25
Justinbarry/ch12725/transactions on tx page displays out of order

justinbarry:


src/modules/portfolio/components/transactions/transactions.jsx
```diff
@@ -105,13 +107,19 @@ export default class Transactions extends Component {
             <Dropdown default={s.transactionPeriodDefault} options={s.transactionPeriodOptions} onChange={this.changeTransactionDropdown} />
           </div>
         </div>
-        { transactions.length === 0 &&
+
+        { transactionsLoading === true &&
+        <div className={PortfolioStyles.Loading__container} >
+          <span>Loading...</span>
```
JohnDanz: not now, but for post MVP we could probably re-used the animated augur-logo here as well like we do on the top left, as a loading indicator.

PR #1896, opened 2018-06-25
Fix calculation of position average price

adrake33:
[CH issue](https://app.clubhouse.io/augur/story/12758/incorrect-position-price)


adrake33:
John also reviewed this and said it should be fine.

PR #1897, opened 2018-06-26
Separate code for creating/verifying liquidity into helper functions

adrake33:
[CH issue](https://app.clubhouse.io/augur/story/12901/create-functions-for-creating-verifying-liquidity-when-creating-markets)


bthaile:
categorical market gets created but no liquidity.

phoebemirman:
i get same outcome as tom

PR #1903, opened 2018-06-26
Tomhaile/ch12932/can progress in create market form without

bthaile:
(create market validation tickSize issue)[https://app.clubhouse.io/augur/story/12932/can-progress-in-create-market-form-without-ticksize-precision]


bthaile:
Does it still happen after code changes

phoebemirman:
yeah

bthaile:
Should be gtg

PR #1910, opened 2018-06-27
augur.js@4.12.0-13

epheph:


epheph:
test failed cause race when i merged (after i saw the actual tests passed)

PR #1913, opened 2018-06-27
Phoebemirman/ch12677/create integration test for disputing walkthrough

phoebemirman:
- first two tests most likely won't pass without a fresh env (assumes no rep for account 2)
- removed some empty it block tests cause those are now being tests in disputing mechanisms section with new verification function


bthaile:
I get this failure even after restarting node and waiting for augur-node to fully synch.

screen shot 2018-06-27 at 10 38 02 pm

PR #1943, opened 2018-07-05
phoebemirman/ch13264/standardize-case-for-tags-cateogries

phoebemirman:
https://app.clubhouse.io/augur/story/13264/standardize-case-for-tags-cateogries


phoebemirman:
Merged by mistake. Need to rework this.

PR #1968, opened 2018-07-08
ClosePositionFixes

JohnDanz:
https://app.clubhouse.io/augur/story/13507/investigate-net-postion-complete-sets


bthaile:
it works like it need to. We need to add the close complete sets button.

PR #1975, opened 2018-07-09
flag to enable mainnet in build

pgebheim:


pgebheim:
Enabling mainnet for next ui build

PR #1994, opened 2018-07-11
johndanz/ch13561/issue-with-pl-s-not-looking-correct

JohnDanz:
https://app.clubhouse.io/augur/story/13561/issue-with-pl-s-not-looking-correct

fixed %s and values to take into account short/long positions and shares being sent.


bthaile:
screen shot 2018-07-11 at 8 35 07 pm


looks like I can loose more than I'm trading. This trade is for multiple orders on the book.

JohnDanz:
The 2 screenshots above actually look correct to me. You can lose more than it costs you because we are talking about the position you will gain and it's potential max profit and loss. So from the examples above:

SELL 0.0061 shares @ .19 = 1 - .19 = .81 * 0.0061 = ~0.00494

BUY 0.006 shares @ .4 = .4 * 0.006 = 0.0024

PR #2044, opened 2018-07-19
Phoebemirman/ch13321/error state for migrating more than account

phoebemirman:
https://app.clubhouse.io/augur/story/13321/error-state-for-migrating-more-than-account-s-rep


bthaile:
Not sure if this is above and beyond. Message still shows up when REP was sent to account while user was on form.

screen shot 2018-07-19 at 10 52 59 pm

PR #2055, opened 2018-07-20
johndanz/ch13775/github-need-ability-to-cancel-order-when

JohnDanz:
https://app.clubhouse.io/augur/story/13775/github-need-ability-to-cancel-order-when-adding-initial-liquidity

I'd like to get either @matt-bullock or @sharkcrayon to get a look at this before merge just to make sure the design/ux/ui looks like they envisioned.


bthaile:
screen shot 2018-07-20 at 7 51 06 pm


Sorting of the orders. Do we want them to be sorted like they will show up in Market Trading view?
Or is the sort as they were entered ok?

matt-bullock:
createmarket_9 5 cancel orders

Looking at Tom's screenshot I would add a little more spacing between the buttons and initial liquidity (56px).

The outcomes could be quite long so I would bunch up the other columns that require less space to make room here. I've attached a screenshot and updated the designs in zeplin.

Is it worth asking the users how they would find the sorting most useful? I think as I add an order I would like to see it appear at the top of the list - so I can see it's there and easily cancel this last order.

sharkcrayon:
I agree with Matt that the user's order of entry should determine the order of the orders in the table.

However, I'm questioning the hierarchy of information here. What is more important to the market creator at this point: seeing the market depth chart and order book, or seeing the list of orders that they've added?

It may be more useful to the user to have the list of orders directly to the right of the order form, and the market depth chart and order book where the order list is now. Or, the market depth chart and order book could become visible when a "Preview" button/tab is clicked.

I'll check with Chwy to see his thoughts on this.

sharkcrayon:
Also, we can take "Share(s)" out of the Quantity column and "ETH" out of the other columns, no? We're not doing that anywhere else and it's repetitive.

JohnDanz:
@sharkcrayon yeah it's actual cost, idk why i put est eth looking back, there will be no surprises on costs at this point.

-- can take out labels, just put them in as a default.

I agree that we don't need to add sorting (matt and I discussed this at one point) i think leaving it as a big list ordered by entry is fine, i also don't think we need to move the table, below everything is fine imo. Maybe Chwy has a different impression but i think this would be fine personally

Chwy5:
The user's order of entry should determine the order of the orders in the table. I would like to see the ability to sort of needed as well.  I think listing the orders at the bottom of the page is fine as long as a user can see it on the same page as the order entry ticket and doesn't need to scroll down to view it.  I also like showing the Market depth chart and order book, as it serves as another check to make sure a users initial liquidity provision is showing as expected.

PR #2081, opened 2018-07-25
Justinbarry/ch11169/add a warning prior to pushing to master

justinbarry:
Git hooks are not able to get user input. Had to go with an env var instead.

If this is acceptable I'll move it to the other Augur repos.

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -11,7 +11,7 @@ if (require.main === module) {
 
   const FLAGS = JSON.parse(process.env.npm_config_argv).original.filter(arg => arg.indexOf("--") !== -1);
 
-  const shouldFix = FLAGS.indexOf("--fix") !== -1 ? true : false;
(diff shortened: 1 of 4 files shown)


scripts/lint.js
```diff
@@ -1,28 +1,23 @@
...
-const testPath = path.resolve(__dirname, '../test');
+let shouldFix = false;
+if (require.main === module) {
+  process.env.FORCE_COLOR = true;
 
-process.env.FORCE_COLOR = true;
+  const FLAGS = JSON.parse(process.env.npm_config_argv).original.filter(arg => arg.indexOf("--") !== -1);
 
-const FLAGS = JSON.parse(process.env.npm_config_argv).original.filter(arg => arg.indexOf('--') !== -1);
+  const shouldFix = FLAGS.indexOf("--fix") !== -1 ? true : false;
```
JohnDanz: don't think you want the conts here

justinbarry: Good catch. Thanks!

Issue #1998, opened 2018-07-12
Escrow is calculated incorrectly if I have complete sets


burdakovd:
See attached pics for current order book and my balances and the order that I'm attempting to submit.

Order should be using just 2 "NO" shares. But instead at least the UI shows that it wants to escrow 3.2 shares. This is wrong.

It happens when orders get matched. If I decrease limit price to avoid matching, the estimate becomes correct.

![image](https://user-images.githubusercontent.com/5193161/42606782-5014514a-8577-11e8-9ffc-45cc92988707.png)
![image](https://user-images.githubusercontent.com/5193161/42606778-4dc036e8-8577-11e8-994f-f9825c59b445.png)

bthaile:
https://app.clubhouse.io/augur/story/13827/calculated-escrowed-eth-incorrect

JohnDanz:
Hi @burdakovd,

Have you happened to notice this issue still being a problem in the latest version of the augur-app (v1.0.4)? 

I have tested this locally using the latest version but was unable to get a situation where it was incorrectly trying to send more shares than would be required. Here's a screenshot:

<img width="1856" alt="screen shot 2018-07-17 at 11 52 21 am" src="https://user-images.githubusercontent.com/10524630/42832920-5da9b4f4-89b8-11e8-8539-575e26f07123.png">

If you are still experiencing this issue on mainnet, then please respond here.

burdakovd:
OK I have similar issue (1.0.4), it escrows less than it should have.
![image](https://user-images.githubusercontent.com/5193161/42922351-1e49fd4e-8b18-11e8-9362-d6e133ca1833.png)

burdakovd:
FWIW here is what I'm getting from Augur node (screenshot of my code logging number of yes/no shares):

![image](https://user-images.githubusercontent.com/5193161/42922396-4921cc5e-8b18-11e8-8d1f-0c40d2666a45.png)

JohnDanz:
@burdakovd hmm ok i see, so one improvement we are planning is on the open orders we will be adding a value for what you have escrowed in those orders to hopefully make this more clear. so maybe that order to sell 2 shares at .78 would also show `2 shares` or `0.44 ETH` depending on what was used to place the order.  Do you have any other suggests to make this less confusing/questionable from a UI perspective?

burdakovd:
It is hard to come up with a good design. If it was up to me, I'd rather make it clear that you have YES/NO shares, and showed their count.

[...]

PR #2083, opened 2018-07-25
johndanz/ch13827/github-add-escrowed-eth-and-shares-to-order

JohnDanz:
https://app.clubhouse.io/augur/story/13827/github-add-escrowed-eth-and-shares-to-order-rows


justinbarry:
Ref #1998

PR #2098, opened 2018-07-27
Claim button in Portfolio: Reporting doesn't fully update after claiming

adrake33:


adrake33:
@phoebemirman I think we can probably hold off on adding Coveralls coverage for this until we have integration tests for claiming reporting fees.

adrake33:
I created https://app.clubhouse.io/augur/story/14632/create-integration-tests-for-claiming-rep-eth to track adding test coverage from claiming REP/ETH as multiple transactions, so I will go ahead and merge this so that this bug can be fixed.

PR #2121, opened 2018-07-31
add notifications for user account to localstorage

bthaile:
https://app.clubhouse.io/augur/story/14588/store-notifications-in-localstorage-until-users-dismisses


bthaile:
not sure how combat that scenario. the messages that signal the tx was confirmed won't come in when the user starts up/refreshes the browser.

PR #2133, opened 2018-08-02
phoebemirman/ch14891/adjust-number-of-decimals-shown-for-eth-and

phoebemirman:
https://app.clubhouse.io/augur/story/14891/adjust-number-of-decimals-shown-for-eth-and-rep-in-the-portfolio-reporting-claim-section

### DIFF
--- a/src/modules/account/components/account-header/account-header.jsx
+++ b/src/modules/account/components/account-header/account-header.jsx
@@ -6,8 +6,8 @@ import Styles from 'modules/account/components/account-header/account-header.sty
 
 const AccountHeader = (p) => {
   // assign defaults incase we have nulls for value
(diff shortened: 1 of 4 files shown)


bthaile:
I don't think we want to round in the core stats. 

User has Rep:   0.99998
screen shot 2018-08-02 at 2 00 43 pm


I would expect to see 0.99998


because of this



screen shot 2018-08-02 at 2 01 34 pm

bthaile:
Also when user has `Rep:   0.000005` the UI says 0.0000

bthaile:
Getting this error trying to trade scalar market.

screen shot 2018-08-02 at 2 08 14 pm

PR #2136, opened 2018-08-03
fix resloved market sorting

bthaile:
fixed sorting resolved markets.

Sort should be 
markets `awaiting_finalization` then `endTime.timestamp`
then `finalized` then `endTime.timestamp`


JohnDanz:
seems fine, just waiting on travis after merging master

PR #2151, opened 2018-08-08
Phoebemirman/ch14995/polish orderbook styles

phoebemirman:
https://app.clubhouse.io/augur/story/14995/polish-orderbook-styles

shows up in create market and trading page


bthaile:
There are 0s when it's tickSize is 0.1 so the prices can only have .1 precision. need to take tickSize (precision) into account. 

screen shot 2018-08-08 at 7 44 56 pm

bthaile:
Just another case of scalar market that has tickSize 1, so there won't be following zeros. Scalars are tricky. :/

screen shot 2018-08-08 at 7 48 02 pm

PR #2160, opened 2018-08-10
add filter and sort options to state to save selected values

phoebemirman:
https://app.clubhouse.io/augur/story/15171/search-filter-settings-should-be-saved-when-navigating-around-the-app-going-back-etc


bthaile:
I see the filter and sorting go back to default even though the drop downs keep their state.


screen shot 2018-08-09 at 8 01 18 pm


to repro:

1. from fresh start, select `in reporting` and `creation time`
2. click on reporting
3. click back to markets

notice that the drop downs are correct but the markets shown are in `open` state

PR #2166, opened 2018-08-13
Phoebemirman/ch15224/my markets in reporting should use dispute

phoebemirman:
https://app.clubhouse.io/augur/story/15224/my-markets-in-reporting-should-use-dispute-version-of-market-card

markets in reporting: show reporting
markets in disputing: show disputing outcomes


matt-bullock:
This shouldn't have a PR yet. It needs design feedback before development can be finished.

PR #2167, opened 2018-08-14
needed to add comparison to markets in component b/c of how outstandi…

bthaile:
…ngReturns is added by action and not originally part of the market object, addes a little bit of clean up

https://app.clubhouse.io/augur/story/15283/claim-and-calculate-payout-buttons-don-t-show-up-on-positions-page-on-initial-load


bthaile:
changes made to fix issue where waiting period pie not showing up.

Issue #2172, opened 2018-08-14
augur-ui installing with errors


kevinreilly:
I can't for the life of me get augur-ui installed locally on OSX. Following the [UI only instructions](https://github.com/AugurProject/augur-ui/blob/master/docs/ui-dev-easy-start.md), with the latest Node, NPM, and Yarn installed with Homebrew as well as Docker and Xcode, when I run `yarn`, I get the output below. I'm mostly a frontend dev so I'm not that great at debugging this kind of stuff. Hopefully someone out there can help me out so I can start contributing to this repo.
```
osx:augur-ui-5.3.0 kevin$ yarn
yarn install v1.9.4
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[----------------------------------------------------------------------] 0/1991(node:32020) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[4/5] 🔗  Linking dependencies...
warning "augur.js > augur-core > solidity-coverage > ethereumjs-testrpc-sc > webpack-cli@2.1.5" has incorrect peer dependency "webpack@^4.0.0".
warning "react-dates > react-with-styles@2.4.0" has unmet peer dependency "react-with-direction@^1.1.0".
warning " > babel-plugin-rewire@1.1.0" has unmet peer dependency "babel-template@^6.2.0".
warning " > babel-plugin-rewire@1.1.0" has unmet peer dependency "babel-types@^6.2.0".
warning " > eslint-import-resolver-babel-module@3.0.0" has incorrect peer dependency "babel-plugin-module-resolver@^2.5.0".
warning "stylelint > postcss-html@0.11.0" has unmet peer dependency "postcss-sass@>=0.2.0".
warning " > stylelint-config-standard@16.0.0" has incorrect peer dependency "stylelint@^7.8.0".
[5/5] 📃  Building fresh packages...
[11/13] ⠈ puppeteer
[8/13] ⠈ node-hid
[-/13] ⠈ waiting...
[9/13] ⠈ secp256k1
[...]

PR #2174, opened 2018-08-15
document the versions of npm, node and yarn we use for development

bthaile:
https://github.com/AugurProject/augur-ui/issues/2172

recommendation by community member that issues building and running augur-ui.


justinbarry:
@bthaile We can programmatically enforce the node version. See: https://docs.npmjs.com/files/package.json#engines

PR #2175, opened 2018-08-15
use hours instead of days for comparison

phoebemirman:
https://app.clubhouse.io/augur/story/14228/dispute-header-progress-bar-should-have-smaller-increments


bthaile:
I would assume the 1 day would transition to hours, not say 0 days. In the example below the fee window has 22 hours left.

screen shot 2018-08-15 at 12 35 36 pm

Issue #2193, opened 2018-08-16
Market tag separator implies hierarchy


kevinreilly:
I propose we change the separator in between market tags from a `/` to a `,`. The tags currently look like they are hierarchical when they are actually siblings.
Current separator:
![augur-tags--old](https://user-images.githubusercontent.com/1888316/44223607-ee203d00-a13c-11e8-9bfd-c8d296435390.jpg)
Proposed separator:
![augur-tags--new](https://user-images.githubusercontent.com/1888316/44223632-0001e000-a13d-11e8-8567-486e405d7d71.jpg)

bthaile:
Discussing this. 

https://app.clubhouse.io/augur/story/15534

sharkcrayon:
@matt-bullock Opinions on this? I agree that the slash may be implying a hierarchy that's not there. But I'm not a fan of the comma, at least with the current design.

bthaile:
How about a dash -

from: TAGS / BITCOIN / BTC
to: TAGS / BITCOIN - BTC

or I like a dot, I dig the dot!

to: TAGS / BITCOIN ∙ BTC

matt-bullock:
Tags need more thought. There are better ways to visually improve the hierarchy here. 

Do we really need to use the word 'Tags' and 'Category'? We could use visual indicators to show they are tags or categories. Could be icons, could be the way they are styled. This results in less clutter.

I'd rather not do a quick change here as there should be a correlation between the way the tags and categories are shown in the menu and how they are shown on the market cards. And of course our whole menu design needs some rethinking.

adrake33:
Issue moved to [AugurProject/augur #309](https://github.com/AugurProject/augur/issues/309) via [**ZenHub**](https://www.zenhub.com/)

PR #2194, opened 2018-08-16
changed market tag separator to comma

kevinreilly:
#2193


bthaile:
This would have to go through design ppl. We might be refactoring navigation, not sure the road map.

bthaile:
closing until we can get design input.

PR #2205, opened 2018-08-19
Fixes the condition on whether to enable https

zchn:
IIUC we want to redirect to the https pages only if $fe_https is on?

### DIFF
--- a/support/nginx-default.conf
+++ b/support/nginx-default.conf
@@ -23,7 +23,7 @@ server {
   listen 80 default_server;
   access_log /var/log/nginx/access.log json_custom;
 
-  if ($fe_https = 'off') {
+  if ($fe_https = 'on') {
     return 302 https://$host$request_uri;
   }
 


pgebheim:
Hello!

This Dockerfile is used to run our dev.augur.net instances. Currently we have a load balancer in front of a 3 instances of the UI, this LB makes http calls to it's backend and forwards the protocol using an X header. 

$fe_https tells us whether or not the front end request came in with https origally, hence the condition is currently correct -- as it will force redirect to https if the frontend didn't not come in with an https connection.

If this docker image is needed more widely, then it could be configurable via environment variable whether to do this redirect in HTTPS.

I would be open to a patch that gives that flexibility, but the one proposed would break our current test infrastructure.

Please let me know what you're looking to do!

zchn:
Ah, thanks for the explanation! I misunderstood the purpose of $fe_https.

For background, I'm currently just playing around with the Augur code base and trying to set up a local Augur instance. My thinking is that if I deploy augur-core onto Ganache and have augur-node and augur-ui pointing to it, it should work. But I haven't got it fully working :(

This PR (as well as the other two) is just something I (mistakenly) discovered along the way.

PR #2237, opened 2018-09-01
make market volume default sort

bthaile:
convo in discord by Mr krug. I agree volume is more interesting.

```
Market default sort should be volume imo(edited)
Just noticed that got changed away to be recently traded
Makes it way harder to consolidate liquidity on open markets which is important for the success of this imo
Dunno how that got changed
```


JohnDanz:
Going to close this as Joey has said Open Interest is fine.

PR #2251, opened 2018-09-05
Write integration tests for P/L

adrake33:
[CH task](https://app.clubhouse.io/augur/story/14392/write-integration-test-for-p-l)

I talked with Paul about what should be tested for P/L, and he recommended the following:

(1) Buying a share, and then another account doing a buy at a different price will change the unrealized PL
(2) Buying and selling at a different price will change the realized PL
(3) Doing (1) then (2) will affect both the Unrealized and Realized PL

He also said that buying NO by selling YES on a binary market will incorrectly affect the PL because we don't give any initial cost to the NO shares. (This is a bug.) The second integration test tests this scenario, but I think the P/L in this case is also affected by a bug that displays small negative numbers as "-0.000" on the trading page (https://app.clubhouse.io/augur/story/15811/fix-p-l-display-for-small-negative-numbers). These tests should be updated once this bug is fixed.

### DIFF
--- a/integration/helpers/dismiss-disclaimer-modal.ts
+++ b/integration/helpers/dismiss-disclaimer-modal.ts
@@ -6,7 +6,7 @@ export const dismissDisclaimerModal = async (page: Page) => {
   );
   if (isModalDismissed) return;
 
-  const timeoutMilliseconds = 5000;
+  const timeoutMilliseconds = 10000;
   await expect(page).toClick(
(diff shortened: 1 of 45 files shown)


adrake33:
Using the toMarket function was causing some problems. I'm still not quite sure why it wasn't working. (I could get it to work in the first test, but not the second.) For now, I reverted my previous change and have the tests going to the Market page by clicking. I also made a few minor changes that were causing the tests to occasionally fail.

PR #2257, opened 2018-09-06
Juliyasmith/ch15719/update search params when connect network

unparalleled-js:
https://app.clubhouse.io/augur/story/15719/update-search-params-when-connect-network-modal-form-is-submitted


src/modules/modal/components/modal-network-connect/modal-network-connect.jsx
```diff
@@ -9,21 +9,8 @@ import {
 import Input from "modules/common/components/input/input";
 
 import Styles from "modules/modal/components/modal-network-connect/modal-network-connect.styles";
-
-function calculateConnectionErrors(err, res) {
```
JohnDanz: Did we intend to remove the error messaging? Now we appear to do a full page refresh when we click "connect" when previously it would retry the new connection and display an error if it failed. This may be intentional, just don't see it mentioned in the ticket so i want to make sure.

PR #2265, opened 2018-09-08
added padding-bottom to allow for market-list to show full card with …

bthaile:
…space when showPagination is false

fix market-list cutting off the bottom when showpagination is false, fix padding-bottom for null state, needs more padding now.


bthaile:
I see an improvement that can be made. Closing for now.

PR #2270, opened 2018-09-10
phoebemirman/ch15224/my-markets-in-reporting-should-use-dispute

phoebemirman:
https://app.clubhouse.io/augur/story/15224/my-markets-in-reporting-should-use-dispute-version-of-market-card


bthaile:
screen shot 2018-09-10 at 9 05 05 pm


seeing market `in dispute` but the `view` button is shown.

bthaile:
Its confusing. In to other places we have ‘in dispute’ sections and the user can dispute. We’ll wait for the community to decide then.

PR #2277, opened 2018-09-11
Fix reporting test & break into smaller tests

adrake33:
[CH issue](https://app.clubhouse.io/augur/story/15931/fix-reporting-reporting-integration-test)

Tests were broken up into smaller tests because the original test took a long time to run, and this allows for faster iteration if a single test fails.


adrake33:
@unparalleled-js Market creation still seems to be inconsistent. It looks like markets aren't always getting created via flash as expected for some reason. I'll investigate this further and let you know when these tests are passing more consistently.

adrake33:
@unparalleled-js I made some changes to these tests and ran them individually. They're passing much more reliably now.

unparalleled-js:
> @unparalleled-js I made some changes to these tests and ran them individually. They're passing much more reliably now.

Cool! I will have time to review this tonight.

adrake33:
Reopening this, as I meant to close a different PR.

PR #2286, opened 2018-09-13
Augur ui refactor 2nd pass

JohnDanz:
Removed unused code, combined up some single key/values that were in state such as `isLogged` into containing object. 3 new top level state objects that contain old top level state booleans/strings
```
appStatus (hasLoadedMarkets, isAnimating, isMobile, isMobileSmall, transactionsLoading)
authStatus (isLogged, ledgerStatus, edgeContext, edgeLoading)
filterSortOptions (marketFilter, marketSort, transactionPeriod)
```
combined/removed some actions/reducers around those values for single action/reducers for updating/managing the new top level states. Combined some more orders/markets actions/reducers where it made sense to. Goal was to reduce file count where it made sense.


src/modules/markets-list/containers/markets-view-container.js
```diff
@@ -29,8 +29,8 @@ const mapStateToProps = (state, { location }) => {
     isMobile: state.appStatus.isMobile,
     markets,
     category,
-    defaultFilter: state.filterOption,
-    defaultSort: state.sortOption
+    defaultFilter: state.filterSortOptions.marketFilter,
```
justinbarry: Selector?

src/modules/account/containers/account-withdraw.js
```diff
@@ -10,7 +10,7 @@ const mapStateToProps = state => {
   return {
     eth: loginAccount.eth,
     rep: loginAccount.rep,
-    isMobileSmall: state.isMobileSmall
+    isMobileSmall: state.appStatus.isMobileSmall
```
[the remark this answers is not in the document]
bthaile: I think that is ok, no logic is needed. just property

src/modules/markets-list/containers/markets-view-container.js
justinbarry: The complexity of the selector isn't the only consideration. Decoupling the shape of state from the components is also something to consider.

PR #2300, opened 2018-09-17
johndanz/ch11899/use-new-ledger-libs-that-don-t-dep-on-node

JohnDanz:
https://app.clubhouse.io/augur/story/11899/use-new-ledger-libs-that-don-t-dep-on-node-hid

Adjusted package.json to allow for node 10.
removed ethereumjs-ledger (rip) for the ledgerjs lib
Adjusted approval/fillOrderOnly detection/notification text


justinbarry:
Nice!

PR #2302, opened 2018-09-19
Tomhaile/ch16626/logged in user s eth balance doesn t update

bthaile:
https://app.clubhouse.io/augur/story/16626/logged-in-user-s-eth-balance-doesn-t-update-when-eth-is-transferred

make sure users ETH balance is update on each block.


justinbarry:
LGTM

justinbarry:
@bthaile I rebased this branch onto master to cleanup the history.

PR #2320, opened 2018-09-24
Justinbarry/ch16863/upgrade to webpack 4

justinbarry:
https://app.clubhouse.io/augur/story/16863/upgrade-to-webpack-4


.gitignore
```diff
@@ -13,6 +13,8 @@ target
 scratch
 .vscode
 
+package-lock.json
```
JohnDanz: probably should add yarn.lock as well if we are going to start ignoring lock files

justinbarry:
@JohnDanz I'm going to remove that commit. It is really outside the scope of this ticket anyway.

Thank you for looking at it.

PR #2321, opened 2018-09-24
Achapman/ch16781/chart displays weird for small number on

phoebemirman:
https://app.clubhouse.io/augur/story/16781/chart-displays-weird-for-small-number-on-scalar


src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
```diff
@@ -161,11 +163,18 @@ export default class CreateMarketOutcome extends Component {
     updateNewMarket(updatedMarket);
   }
 
-  validateScalarNum(value, type) {
+  validateScalarNum(rawValue, type, limitDigits) {
```
bthaile: seems we need this function in format-number.

PR #2325, opened 2018-09-25
Juliyasmith/ch10283/refactor range display

unparalleled-js:
https://app.clubhouse.io/augur/story/10283/refactor-range-display


unparalleled-js:
Everything is working besides the market-indicator is missing from the extreme right-sided case

PR #2337, opened 2018-09-26
auto login user when they change MM accounts

bthaile:
https://app.clubhouse.io/augur/story/16973/auto-login-user-when-they-switch-mm-accounts

auto login user when they chose different MM account.

### DIFF
--- a/src/modules/app/actions/init-augur.js
+++ b/src/modules/app/actions/init-augur.js
@@ -96,12 +96,19 @@ function loadAccount(dispatch, existing, accountType, callback) {
         usingMetaMask &&
         loggedInAccount !== account
       ) {
+        // local storage does not match mm account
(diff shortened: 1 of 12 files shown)


src/modules/app/actions/init-augur.js
```diff
@@ -97,8 +97,8 @@ function loadAccount(dispatch, existing, accountType, callback) {
         loggedInAccount !== account
       ) {
         // local storage does not match mm account
-        dispatch(logout());
-        account = null;
+        dispatch(useUnlockedAccount(account));
+        loggedInAccount = account;
```
phoebemirman: also, for hosted environment it still logs you out of ui when you switch mm accounts. need to switch logout func in next few if/else statement as well

PR #2344, opened 2018-09-27
changed to not do gas estimate if user doesn't have enough balance fo…

bthaile:
…r validity and no-show bond, also added in liquidity when calculating if user has sufficient ETH

https://app.clubhouse.io/augur/story/17032/create-market-check-user-balance-before-getting-gas-estimates

verify user gets notified if they don't have enough ETH and REP to create market (gas cost is blank)
verify that gas cost shows up when they do have enough ETH and REP but don't have enough to also pay gas
verify that user is notified that they don't have enough ETH to cover liquidity.

### DIFF
--- a/src/modules/create-market/components/create-market-form-review/create-market-form-review.jsx
+++ b/src/modules/create-market/components/create-market-form-review/create-market-form-review.jsx
@@ -63,7 +63,7 @@ export default class CreateMarketReview extends Component {
     this.calculateMarketCreationCosts();
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps(nextProps, nextState) {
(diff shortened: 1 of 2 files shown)


AugurIntegration:
This pull request **fixes 1 alert** when merging 43851a50c24c6c6e22d6ceb10355ba07a2d7237d into bd7eb0a6134d8e3b2a6a23010711dfa608612518 - [view on LGTM.com](https://lgtm.com/projects/g/AugurProject/augur-ui/rev/pr-f292f8158cf2764d2e32ecf9b53c31dfd1b8bbeb)

**fixed alerts:**

* 1 for Unsupported state update in lifecycle method

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #2347, opened 2018-09-28
Juliyasmith/ch16979/these arrow indicators spacing is off

unparalleled-js:
https://app.clubhouse.io/augur/story/16979/these-arrow-indicators-spacing-is-off

### DIFF
--- a/src/modules/market/components/market-outcomes-yes-no-scalar/market-outcomes-yes-no-scalar.jsx
+++ b/src/modules/market/components/market-outcomes-yes-no-scalar/market-outcomes-yes-no-scalar.jsx
@@ -53,7 +53,7 @@ const MarketOutcomes = p => {
       window.outerWidth < 590
         ? { marginTop: 4, fontSize: 16 }
         : { marginTop: 7, fontSize: 20 };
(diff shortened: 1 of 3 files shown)


src/modules/market/components/market-outcomes-yes-no-scalar/market-outcomes-yes-no-scalar.jsx
```diff
@@ -61,10 +44,23 @@ const MarketOutcomes = p => {
...
+    const isMobileAttrs =
+      window.outerWidth < 590
+        ? { marginTop: 4, fontSize: 16 }
+        : { marginTop: 7, fontSize: 20 };
+    for (let i = 6; i < size; i += 6) {
+      size *= 0.7;
+    }
+    return {
+      marginLeft: pos < 15 ? size + "em" : 0,
+      marginRight: pos > 85 ? size + "em" : 0,
```
adrake33: Can these be changed to use rem instead of em? I believe we're only supposed to be using rem.

unparalleled-js: done!

PR #2362, opened 2018-10-02
Associate notifications with a network ID & universe

adrake33:
For old notifications (where the network ID & universe are not set), this code will set the network ID & universe to whatever network/universe the user is using when these old notifications are first encountered. Since most users do not switch networks often and are usually on mainnet, this means that, in most cases, their old notifications will get associated with mainnet. This may not always be the desired behavior, but we currently have no way of guaranteeing which network old notifications should be associated with.


bthaile:
as we already discussed, just want to put in a reminder note. We need to not filter when localStorage is pulled in but at the viz layer, filter on universe and networkId. Show notifications that have neither.

adrake33:
@bthaile I think this should be ready now. I made one more minor improvement to the code, and it seems to be working as expected when switching between users/networks and doing hard refreshes. Please test it out some more and merge if everything looks good.

PR #2363, opened 2018-10-02
Corrected spelling of pagination

unparalleled-js:


src/modules/portfolio/components/positions/positions.jsx
```diff
@@ -54,7 +54,7 @@ export default class Positions extends Component {
               history={history}
               currentTimestamp={currentTimestamp}
               isMobile={isMobile}
-              pageinationName="open"
+              MockAugurJSName="open"
```
phoebemirman: is this name a mistake?

unparalleled-js: must be, just fixed that. whoops.

PR #2396, opened 2018-10-13
Tomhaile/ch17466/categorical market weird issue with volume

bthaile:
https://app.clubhouse.io/augur/story/17466/categorical-market-weird-issue-with-volume-gte-big-number-conditional


adrake33:
Tom is going to create a better fix for this by modifying `createBigNumber` to always convert the first param to a string if it's a number. Closing this PR.

PR #2400, opened 2018-10-17
johndanz/ch17343/ledger-trezor-address-paging-need-transition

JohnDanz:
added an animation to the address rows to indicate that they are loading, the rows will now pulse/fade in and out until an address/balance is loaded.


bthaile:
Need same pulsing behavior when paging forward and backward.

PR #2405, opened 2018-10-19
phoebemirman/ch17541/general-knowledge-resolution-source-is-missing

phoebemirman:
https://app.clubhouse.io/augur/story/17541/general-knowledge-resolution-source-is-missing

screen shot 2018-10-19 at 11 06 50 am


adrake33:
@bthaile The Resolution Source was not being displayed when it was set to General Knowledge. Should this be fixed in a different way?

PR #2433, opened 2018-10-26
Phoebemirman/ch17599/errors thrown when forking market is finalized

phoebemirman:
https://app.clubhouse.io/augur/story/17599/errors-thrown-when-forking-market-is-finalized


src/modules/market/components/core-properties/core-properties.jsx
```diff
@@ -152,7 +152,7 @@ export default class CoreProperties extends Component {
           value: getValue(market, "volume.full")
         },
         {
-          name: "fee",
+          name: "est. fee",
```
bthaile: This value is the actual fee percentage, not estimate.

PR #2456, opened 2018-11-05
Justinbarry/ch17919/upgrade ui to react 16 3 or greater

justinbarry:
This is required to use the 0x widget. It seems to work but I would like both of your opinion.


justinbarry:
@phoebemirman Thanks! Turns out this isn't necessary.

PR #2468, opened 2018-11-07
fix-big-number-math-with-js-numbers

JohnDanz:
modified createBigNumber to cast the value passed to a string if its a JS Number

### DIFF
--- a/src/utils/create-big-number.js
+++ b/src/utils/create-big-number.js
@@ -4,8 +4,7 @@ import { BigNumber } from "bignumber.js";
 export const createBigNumber = (value, ...args) => {
   let newBigNumber;
   try {
-    const bn = typeof value === "number" ? value.toString() : value;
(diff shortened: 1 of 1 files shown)


src/utils/create-big-number.js
```diff
@@ -1,10 +1,11 @@
 import logError from "utils/log-error";
 import { BigNumber } from "bignumber.js";
 
-export const createBigNumber = (...args) => {
+export const createBigNumber = (value, ...args) => {
   let newBigNumber;
   try {
-    newBigNumber = new BigNumber(...args);
+    const bn = typeof value === "number" ? value.toString() : value;
```
justinbarry: Should it ever be anything but a string? If not, try this?
```suggestion
    const bn = `${value}`;
```

JohnDanz: good point, will update

PR #2476, opened 2018-11-09
Transactions page should display maker 

phoebemirman:
Story details: https://app.clubhouse.io/augur/story/18012

### DIFF
--- a/src/modules/transactions/actions/add-transactions.js
+++ b/src/modules/transactions/actions/add-transactions.js
@@ -94,7 +94,7 @@ function buildTradeTransaction(trade, marketsData) {
   transaction.id = `${transaction.transactionHash}-${transaction.orderId}`;
   const header = buildHeader(transaction, TRADE, SUCCESS);
   const meta = {};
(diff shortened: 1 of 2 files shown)


src/modules/transactions/actions/add-transactions.js
```diff
@@ -85,7 +86,12 @@ function buildTradeTransactionGroup(group, marketsData) {
...
     // own order filled
-    header.message = formatTransactionMessage(sumBuy, sumSell, "Order", true);
+    header.message = formatTransactioFillMessage(
+      header.transactions[0].type,
+      header.transactions[0].amount,
+      header.transactions[0].price,
+      header.transactions[0].meta && header.transactions[0].meta.outcome
+    );
   } else {
     header.message = formatTransactionMessage(sumBuy, sumSell, "Trade");
```
adrake33: I think we should be calling `formatTransactionFillMessage` here as well. The text should be the same regardless of whether the person filling the trade is filling their own order or someone else's.

PR #2477, opened 2018-11-09
Migrate to 0x instant

justinbarry:


src/modules/account/containers/account-deposit.js
```diff
@@ -1,11 +1,36 @@
...
+      const assetData = assetDataUtils.encodeERC20AssetData(repTokenAddress);
+
+      // eslint-disable-next-line
+      zeroExInstant.render(
+        {
+          availableAssetDatas: [],
+          defaultSelectedAssetData: assetData,
+          orderSource: "https://api.radarrelay.com/0x/v2"
+        },
+        "body"
```
fragosti: this should probably be "#app"

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -141,9 +61,11 @@ export default class AccountDeposit extends Component {
         </div>
         <div className={Styles.AccountDeposit__main}>
           <div className={Styles.AccountDeposit__description}>
-            {shapeShiftConverter}
-            <div />
-            {airSwapConverter}
+            <div className={Styles.AccountDeposit__0xInstantButton}>
+              <button onClick={openZeroExInstant}>
+                Buy REP using 0x instant.
```
fragosti: Just making sure "instant" is capitalized. Is that done via CSS?

justinbarry: Yes.

grahamperich:
I'm curious about why the AirSwap widget was removed here. Should we not allow both of these decentralized atomic swap solutions (0x and AirSwap) and let the user choose whichever method they prefer? 

Would you accept a PR that puts the AirSwap widget back into the UI?

bthaile:
Yes, we would accept a PR for adding back AirSwap.

PR #2493, opened 2018-11-20
Add DeadCodePlugin and remove dead code

ryanberckmans:
See commit msgs


src/modules/create-market/components/create-market-review.jsx
```diff
@@ -1,168 +0,0 @@
...
-import { augur } from "services/augurjs";
-import { BigNumber } from "utils/create-big-number";
-
-import { formatEtherEstimate } from "utils/format-number";
-import getValue from "utils/get-value";
-
-import newMarketCreationOrder from "modules/markets/constants/new-market-creation-order";
-import { NEW_MARKET_REVIEW } from "modules/markets/constants/new-market-creation-steps";
-
-export default class CreateMarketReview extends Component {
```
bthaile: This is used in `src/modules/create-market/components/create-market-form/create-market-form.jsx` so I'm not sure all these files are unused.

ryanberckmans: `import CreateMarketReview from "src/modules/create-market/components/create-market-review.jsx"` I think is dead.

The non-dead type has a similar name:

`import CreateMarketReview from "modules/create-market/components/create-market-form-review/create-market-form-review";`

PR #2501, opened 2018-11-28
Add market creator fees filter dropdown

justinbarry:


src/modules/filter-sort/components/filter-dropdowns/filter-dropdowns.jsx
```diff
@@ -34,15 +42,28 @@ const filterOptions = [
   { value: MARKET_CLOSED, label: "Resolved" }
 ];
 
+const maxFeesOptions = [
+  { label: "All Markets", value: "1" },
```
phoebemirman: this should be a constant for consistency

phoebemirman:
needs some mobile styling. can be in separate pr

PR #2537, opened 2018-12-06
adjust-x-axis-in-performance-chart

JohnDanz:
attempting to adjust the x-axis labels and also shortened them a little bit in the x axis labeling but not the tooltips. adjusted tooltips to always keep ETH on top/date below it, adjust treshholds for when to flip where the tooltip generates


justinbarry:
@JohnDanz Need some space there.
image

PR #2540, opened 2018-12-07
restrict-0x-instant-non-supported-networks

JohnDanz:
this hides the deposit 0x instant button in favor of some explanation text when we are connected to a network that isn't kovan or mainnet.

closes https://github.com/AugurProject/augur/issues/624


JohnDanz:
screen shot 2018-12-11 at 1 50 28 pm
updated button

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -42,12 +43,12 @@ export default class AccountDeposit extends Component {
...
 
   render() {
-    const { address, openZeroExInstant } = this.props;
+    const { address, openZeroExInstant, augurNodeNetworkId } = this.props;
     const styleQR = {
       height: "auto",
       width: "100%"
     };
-
+    const show0xInstant = [1, 42].includes(augurNodeNetworkId);
```
adrake33: Can you obtain these values from NETWORK_NAMES instead? Recommend not hard-coding them.

PR #2546, opened 2018-12-10
Add more complete sets messaging

JohnDanz:
closes https://github.com/AugurProject/augur/issues/636


bthaile:
Shouldn't the `Sell Complete Sets` be a button? A cool condensed button would be awesome. Underlined text is roughly 1998 style. Are we going retro?

screen shot 2018-12-10 at 10 41 40 pm

sharkcrayon:
@bthaile @JohnDanz When would "Sell Complete Sets" be disabled?

@matt-bullock Thoughts on whether the "Sell Complete Sets" link should be a button? I definitely agree with the argument, but I'm not sure we really have enough room there for that.

JohnDanz:
@sharkcrayon it gets disabled after you click it so that it doesn't allow for multiple queueing up of the sell complete sets transaction. if the user rejects the signature/there is an error then it undisables, if it succeeds then this whole button will disappear/user gets ETH back.

sharkcrayon:
Our signing flow sucks. For future reference, if the user clicks the link, the tx popup gets hidden (say behind the browser window), the link is then disabled and the user doesn't know why. And there's no recourse to reenable it.

@JohnDanz After the user signs, can you make the purple notification bar about having complete sets disappear?

sharkcrayon:
Ok, talked with @JohnDanz more.

Here are the states we want:
- User has complete sets to sell: "Sell Complete Sets", enabled
- User clicks "Sell Complete Sets" but hasn't signed: "Sell Complete Sets", disabled
- User signs the tx: "Pending", not clickable
- Tx succeeds: Purple bar about complete sets disappears
- Tx fails: Purple bar about complete sets remains, "Sell Complete Sets", enabled

matt-bullock:
As it has a disabled state I recommend using this button. Unfortunately we don't have a slimmer button in the current design (we do in the redesign). 
image

this is  the disabled state:
image

AugurIntegration:
This pull request **introduces 2 alerts** when merging fca2002ae4a1aec8159d065c0c69b5fabf67a662 into 0c2b681143145f04592953480bcc8f40ae87c2bb - [view on LGTM.com](https://lgtm.com/projects/g/AugurProject/augur-ui/rev/pr-37fc0b26a46ca9f9a1a2c14156140fd74f6c182f)

**new alerts:**

* 2 for Useless assignment to local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/modules/portfolio/components/market-portfolio-card/market-portfolio-card.jsx
```diff
@@ -112,6 +120,22 @@ export default class MarketPortfolioCard extends Component {
       default:
         localButtonText = "View";
     }
+    const pendingCompleteSetsHash = `pending-${market.id}-${numCompleteSets &&
+      numCompleteSets.fullPrecision}`;
+    const pendingCompleteSetsInfo = transactionsStatus[pendingCompleteSetsHash];
+    const status = pendingCompleteSetsInfo && pendingCompleteSetsInfo.status;
+    let completeSetButtonText = "Sell Complete Sets";
+    switch (status) {
```
bthaile: Minor: this switch stmt comes up twice, it should be in it's own selector. Then we can grow it out for all tx statuses.

src/modules/positions/actions/sell-complete-sets.js
```diff
@@ -18,18 +26,27 @@ export function sellCompleteSets(
...
       numTicks
     );
+    const pendingHash = `pending-${marketId}-${numCompleteSets.fullPrecision}`;
     const sellCompleteSetsParams = {
       tx: {},
       meta: loginAccount.meta,
       _market: marketId,
       _amount: numCompleteSetsOnChain,
-      onSent: res => callback(null, res),
+      onSent: res => {
```
bthaile: 👍

PR #2554, opened 2018-12-12
use height instead of character count

phoebemirman:
fixes https://github.com/AugurProject/augur/issues/688


phoebemirman:
it is

PR #2567, opened 2018-12-21
Update profit loss

JohnDanz:
Needs to be paired with Paul's Merge-PL branch https://github.com/AugurProject/augur-node/tree/pgebheim/merge-pl 

this updates positions/PL calls and selectors to handle the new calculations/endpoints in augur-node.


bthaile:
A few issue just trying to trade scalar market:

"Millions of Tether tokens issued on Sat Nov 17 2018 (round down)"
take all bid orders, and get this profit/loss estimations. Shouldn't my loss be all ETH escrowed?

screen shot 2019-01-04 at 8 38 47 pm


Second issue is I get a position but doesn't show on Market Trading page, no errors. My position does show up on Portfolio Positions page.

example of not showing I have a position but do own shares, they show up when I try to trade them:

screen shot 2019-01-04 at 8 45 22 pm



When I went back to Portfolio Positions page I get an error for performance chart.

screen shot 2019-01-04 at 8 43 20 pm

bthaile:
Also confirmed for both scalar and binary markets the user's position doesn't show up on the market trading page.

Issue #2510, opened 2018-11-30
Fill order notification issues


adrake33:
Reported by @Chwy5:

I started with a short position. The best offer on the order book showed 1 share at .8 (which i clicked on, to filter into the order ticket) I then changed the order ticket quantity to 2 shares and had the fill order only button checked....So I was a buyer of 2 shares (knowing i would only be filled on one)

![11-29wrongmessage.jpg](https://images.zenhubusercontent.com/5bff29dc5cb43a45a18add9f/5b60238e-2691-4dcb-9b8e-46f8dcac9cd3)

The message is wrong...it should show buy order and not sell.

Also, checking the fill order box and  changing the quantity - which will result in a partial fill ( i was filled on 1 share and my other didnt hit the book, like expected) resulted in a pending state (even though I have an etherscan confirm)

adrake33:
Issue moved to [AugurProject/augur #255](https://github.com/AugurProject/augur/issues/255) via [**ZenHub**](https://www.zenhub.com/)

Issue #255, opened 2016-01-18
Augur top left link


joeykrug:
Clicking on AUGUR on the top left of the site shouldn't reload the page

tinybike:
Right now this is intended behavior.  Am open to suggestions though!

joeykrug:
With a site that takes ~20 seconds for me to load on Windows 10 Opera having any refreshes beyond one I purposefully do by either F5 or clicking the refresh button is annoying

tinybike:
Point taken.  I'll make it just link back to Overview instead of doing a full refresh.

tinybike:
Added in 7f7e9549adc5d765b0848841a00901a2c12e54c7.

PR #2570, opened 2018-12-21
Fill order notification fix issue #2510

JohnDanz:
potentially a wip


JohnDanz:
Fill order notification issues #255

adrake33:
I went to a market with a sell order for 0.002 shares of YES at 0.35 ETH and placed a buy order for 0.004 shares at 0.35 ETH. I got one notification like:
```
CONFIRMED
Place trade
Fill buying 0.0020 Shares of "Yes" at 0.3500 ETH
 — 2 minutes ago
```
which was expected. However, since there were no other orders on the book that could be filled, the second notification got stuck in pending. When I went to another page, I saw:
```
CONFIRMED
Place trade
Unable to buy Shares of "Yes" at 0.35 ETH.
 — 2 minutes ago
```
Is the expectation that the second order will stay stuck in pending until the user goes to a different page?  Is there a way this can get updated to confirmed status while staying on the page?

adrake33:
I retried the above scenario, and both transactions got confirmed.

PR #2574, opened 2018-12-28
Account deposit update

joeykrug:


src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -92,25 +92,65 @@ export default class AccountDeposit extends Component {
...
+                  Buy ETH (for trading) using Wyre
                 </button>
               </div>
+            }
+            {
+              <>
+                <br />
+                <div className={Styles.AccountDeposit__0xInstantButton}>
+                  <button
+                    onClick={() => window.open("http://www.coinbase.com/")}
```
adrake33: Perhaps this should go to https://www.coinbase.com/buy/ETH instead?

src/modules/account/components/account-deposit/account-deposit.jsx
```diff
@@ -92,25 +92,65 @@ export default class AccountDeposit extends Component {
...
+            {show0xInstant && (
+              <>
+                <br />
+                <div className={Styles.AccountDeposit__0xInstantButton}>
+                  <button onClick={openZeroExInstant}>
+                    Buy REP (for reporting) using 0x instant
+                  </button>
+                </div>
+              </>
             )}
```
adrake33: I think we should add more spacing or have some sort of divider between the buttons for buying ETH and buying REP. This will look a bit strange on mainnet/Koven with all the buttons squished together:

screen shot 2019-01-02 at 4 02 07 pm

src/modules/account/components/account-deposit/account-deposit.jsx
adrake33: Looks like the one for Wyre would be https://dash.sendwyre.com/dashboard

src/modules/account/components/account-deposit/account-deposit.jsx
joeykrug: Fixed

PR #2593, opened 2019-01-07
fixed-proptypes-and-formatting

JohnDanz:
updated proptypes/default props and cleaned up some react formatting


justinbarry:
I know I’m late to the party here but nice catch!

PR #2594, opened 2019-01-08
Create sitemap.xml

joeykrug:


joeykrug:
Added a working sitemap back for dev site for further testing

PR #2601, opened 2019-01-09
Preload fonts

joeykrug:


joeykrug:
Has some junk from a merge from latest + changes to add font preloads

bthaile:
Are these certs suppose be added if so what for?

joeykrug:
No they’re not. Dunno what they are or where they came from haha. Can you remove, on mobile atm

PR #2626, opened 2019-01-18
fix-disclaimer-css-mobile

JohnDanz:
fixed the sizing of the disclaimer modal on mobile after adding the text above the checkbox


adrake33:
I tested this using XCode, and the bottom bar in Safari still blocks the button on the iPhone SE:

screen shot 2019-01-18 at 12 48 58 pm

Also, I'm seeing the Google Translate banner come up by default at the top of the screen, which is pushing the modal down too far even on the iPhone X (probably the result of the recent SEO changes):

screen shot 2019-01-18 at 12 41 21 pm

adrake33:
I would also recommend reducing the margins for the text above the checkbox. It looks really squished together and doesn't need to take up so much space vertically.

adrake33:
Fixes 878 and 886

PR #2628, opened 2019-01-20
Profile UI Adjustments

Brayden967:
Made some simple UI changes to the profile page, attached screenshot below.

Reason for the adjustments:
->Trade buttons went outside of view range on profile page
->QR Code had a dedicated section, which could be better suited under the public address

Changes Made:
->Moved trade buttons to a dedicated column
->Moved QR code to show under public address (makes more sense, as these are the same)
->Changed some wording on the three steps 

adjustments


Brayden967:
@joeykrug Thanks for the feedback Joey, I implemented the changes you had suggested and I also changed some of the wording for the three steps as I feel this might make a bit more sense for new users. This implies that the public address can be used to receive both REP & ETH.

joeykrug:
Awesome this is great, thanks!

PR #2661, opened 2019-02-11
fix-dispute-choices

JohnDanz:
fixed an issue that would cause the 8th outcome to not display on a dispute form if invalid is the tentative winning outcome of an 8 category categorical market.

To test locally, fire up a pop node but not with normal time.

make an 8 outcome categorical market, push to reporting using flash, report `INVALID` then push 1 week to dispute rounds, go to dispute form and notice that outcome 1-8 are all being displayed properly now.


bthaile:
Bice one

PR #2664, opened 2019-02-12
restyle header

phoebemirman:


src/modules/market/components/core-properties/core-properties.styles.less
```diff
@@ -40,6 +40,7 @@
 
   span:nth-of-type(2) {
     color: @color-white;
+    font-family: 'Roboto Mono', monospace;
```
bthaile: Dont we have a constand for font family

src/modules/market/components/market-header/market-header.styles.less
```diff
@@ -230,6 +241,7 @@
 
 .Marketheader__resolutionSource {
   font-family: 'Roboto Mono', monospace;
```
bthaile: Constand for font family

PR #2665, opened 2019-02-12
Report button bug

phoebemirman:
- On Open Orders, change 'Escrowed Eth' and 'Escrowed Shares' labels to 'Total Cost (ETH)' and 'Total Cost (Shares)' 

- On a market that can be reported on, if I go to the trading pg and hit report button nothing happens

### DIFF
--- a/src/modules/market/components/market-header/market-header-reporting.jsx
+++ b/src/modules/market/components/market-header/market-header-reporting.jsx
@@ -8,7 +8,11 @@ import Styles from "modules/market/components/market-header/market-header-report
 import { constants } from "services/constants";
 import MarketLink from "modules/market/components/market-link/market-link";
 import MarketHeaderStyles from "modules/market/components/market-header/market-header.styles";
-import { CATEGORICAL, TYPE_DISPUTE } from "modules/common-elements/constants";
+import {
+  CATEGORICAL,
+  TYPE_DISPUTE,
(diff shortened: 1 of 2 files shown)


src/modules/market/components/market-orders-positions-table/open-orders-table.jsx
```diff
@@ -30,8 +30,8 @@ const OpenOrdersTable = ({
           <span>Quantity</span>
         </li>
         <li>Price</li>
-        <li>Escrowed ETH</li>
-        <li>Escrowed Shares</li>
+        <li>Total Cost (ETH)</li>
```
bthaile: Is this changing to `Frozen Funds`?

src/modules/market/components/market-header/market-header-reporting.jsx
```diff
@@ -252,6 +258,7 @@ export default class MarketHeaderReporting extends Component {
                 id={id}
                 location={location}
                 disabled={!isLogged}
+                linkType={TYPE_REPORT}
```
bthaile: aren't we going to have a standard button to replace MarketLink?

bthaile: nvm, we don't have a new button yet, MarketLink will be refactored when we do.

PR #2667, opened 2019-02-12
display-market-author

JohnDanz:
added market author to the top of the market trading page.

To test, go to a market page and confirm that you can see the market creator address.


yarn.lock
```diff
@@ -905,81 +905,59 @@
     lodash "^4.17.10"
```
andrewdamelio: Should we also update package-lock.json? or do we only maintain yarn locks

PR #2673, opened 2019-02-13
Update Scalar Denomination Validation

JohnDanz:
updated scalar denomination to be required and also to trim the string before checking to make sure we dont allow just spaces which will appear blank in the outcome name


src/modules/create-market/components/create-market-form-outcome/create-market-form-outcome.jsx
```diff
@@ -188,6 +188,8 @@ export default class CreateMarketOutcome extends Component {
         ? value
         : updatedMarket.scalarDenomination;
 
+    console.log(denomination);
```
andrewdamelio: leftover console log

PR #2675, opened 2019-02-14
Add movement label

andrewdamelio:
### Create  Price Movement Label

- Added a common element to use for price movement display. There are quite a few places in the redesign that will use this for display. 

- I Will add this to (and fix other redesign style issues on) the TopBar component, but on a seperate PR.

### DIFF
--- a/src/modules/common-elements/movement-text.tsx
+++ b/src/modules/common-elements/movement-text.tsx
@@ -1,0 +1,86 @@
+import React from "react";
+import classNames from "classnames";
+import * as constants from "./constants";
+import Styles from "./movement-label.styles";
+
+export enum sizeTypes {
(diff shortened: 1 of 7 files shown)


src/modules/common-elements/movement-text.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const addBrackets: Function = (label: string): string => {
+    if (props.showBrackets) {
+      return `(${label})`;
+    }
+    return label;
+  };
+
+  const formattedString = addBrackets(
+    addPercent(addPlus(toString(removeMinus(props.value))))
```
acr13: If possible can we do this in one step? Seems like lots of string manipulation that can be done in one go

PR #2679, opened 2019-02-15
fallback to hosted nodes in ui

andrewdamelio:
#### Test Cases:

- User is connected to nether a local ETHEREUM/AUGUR NODE **(used both fallbacks)**
- User is connected to only a AUGUR NODE **(use eth-node fallback)**
- User is connected to only a ETHEREUM NODE **(use augur fallback)**
- User is connect to both a local ETHEREUM NODE and AUGUR NODE **(ignore fallbacks)**


andrewdamelio:
> Are we going to hardcode augur.casino and alchemyapi. I’m not sure about that.

see https://github.com/AugurProject/augur/issues/882

PR #2695, opened 2019-02-26
Add notification card - Open Orders in Resolved Market

andrewdamelio:
### To Test
**Path 1**
1) Create an open order in a open market
2) Wait until market is finalized
3) You should see notification on Account Summary page

**Path 2**
1) Create an open order in a already finalized market
2) You should see notification on Account Summary page

------
#### TODO (in other work PR's)
- store and load form localStorage
- Mark as read when clicked
- Wire up corresponding Modal


src/modules/account/components/notifications/notifications-templates.tsx
```diff
@@ -0,0 +1,59 @@
...
+    } Please review and cancel these orders to release your ETH.`}
+    marketName={props.marketName}
+  />
+);
+
+// Helper
+const wrapMarketName = (marketName: string) => (
+  <span
+    className={Styles.NotificationCard__MarketName}
+  >{`"${marketName}"`}</span>
```
bthaile: Do we want this market name to be clickable so user can nav to market?

If so use 
```
                <MarketLink id={props.market.id}>
                  {props.market.description}
                </MarketLink>
```

src/modules/notifications/reducers/notifications.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface Action {
+  type: string;
+  data: Data;
+}
+
+export default (notifications = DEFAULT_STATE, action: Action) => {
+  switch (action.type) {
+    case UPDATE_NOTIFICATIONS:
+      return action.data.notifications
```
bthaile: do you want to blow away old notifications when adding them?

src/modules/notifications/selectors/notification-state.ts
```diff
@@ -0,0 +1,46 @@
...
+        .map(marketId => {
+          return {
+            marketId,
+            marketName: marketData[marketId].description
+          }
+        });
+
+      // Get all Open Orders for a provided market
+      const hasOpenOrders = (marketId: string)  => {
+        const market = selectMarket(marketId);
```
bthaile: Instead of going though the markets via marketsData.

You can use `selectMarket` and filter on marketStatus === 'resolved' this will get `finalized` and `awaiting_finalization` markets.

Also don't use `userOpenOrders` on outcomes, that is going away. user `market.userOpenOrders` it'll be faster.

also using these two steps you don't have to filter the markets in the end they are already filtered.

could do something like this, as an example. prob easier with a .map 🤷‍♂️ 
```
  const userOrders = selectMarkets(state).reduce(
    (p, market) => [...p, ...market.userOpenOrders],
    []
  );
```

src/modules/notifications/reducers/notifications.ts
andrewdamelio: For now -- this will change when we incorporate localStorage.
I want to push the heavy reducer logic in a separate PR.

src/modules/account/components/notifications/notifications-templates.tsx
andrewdamelio: Checking with Matt - I think we will make this change just confirming.

src/modules/notifications/selectors/notification-state.ts
andrewdamelio: 👍 💯

PR #2702, opened 2019-02-27
updated validation for quantity on the trade form

JohnDanz:
https://github.com/AugurProject/augur/issues/1024


bthaile:
not going to do this. won't be an issue with DAI, also we have message to user if the gas cost is greater than order value.

bthaile:
We do need this, because order form can take in more than 15 decimals which can crash BigNumber. We can handle this by limiting input.

PR #2707, opened 2019-02-28
load DR markets for currently logged in user

bthaile:
closes https://github.com/AugurProject/augur/issues/1186

loads all currently logged in user's markets if they are designated reporter.


andrewdamelio:
tested, worked perfectly 👍

PR #2716, opened 2019-03-04
consolidated pendingOrders appending to userOpenOrders to market sele…

bthaile:
fixes https://github.com/AugurProject/augur/issues/1202

This is to simplify code in containers. removed a few files that are no longer used.


src/modules/markets/selectors/market.js
```diff
@@ -356,6 +360,11 @@ export function assembleMarket(
             .filter(collection => collection.length !== 0)
             .flat() || [];
 
+        // add pending orders
+        if (pendingOrders && pendingOrders.length > 0) {
+          market.userOpenOrders = market.userOpenOrders.concat(pendingOrders);
```
phoebemirman: should pending orders go first so they show up in the lists first?

bthaile: yes, they should. making change

bthaile: I see your comment and raise you a commit.

PR #2720, opened 2019-03-06
add Sell Complete Sets and Dispute notifications

andrewdamelio:
### Add dispute/sell full shares notifications

@bthaile For showing disputeInfo for position holder users, I had to add ```dispatch(loadDisputing());``` to ```load-account-data.js```. I didn't notice a performance hit with this but please let me know if there is a more performant way to go about this.

### DIFF
--- a/src/modules/account/components/notifications/notifications.tsx
+++ b/src/modules/account/components/notifications/notifications.tsx
@@ -6,6 +6,8 @@ import {
   FinalizeTemplate,
   OpenOrdersResolvedMarketsTemplate,
   ReportEndingSoonTemplate,
+  DisputeTemplate,
+  SellCompleteSetTemplate
 } from "modules/account/components/notifications/notifications-templates";
 
 import { Market, Notifications as INotifications  } from "modules/account/types";
@@ -15,6 +17,8 @@ export interface NotificationsProps {
   resolvedMarketsOpenOrders: Array<Market>;
   reportOnMarkets: Array<Market>;
   finalizedMarkets: Array<Market>;
+  marketsInDispute: Array<Market>;
+  completeSetPositions: Array<Market>;
   updateNotifications: Function;
   notifications: Array<INotifications>;
   currentAugurTimestamp: number;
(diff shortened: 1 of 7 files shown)


src/modules/account/components/notifications/notifications.tsx
```diff
@@ -29,16 +33,20 @@ class Notifications extends React.Component<NotificationsProps> {
...
           this.updateState(nextProps);
     }
   }
 
   updateState(props: NotificationsProps) {
     this.props.updateNotifications(
-      this.generateCards(this.props.resolvedMarketsOpenOrders, 'resolvedMarketsOpenOrders')
-      .concat(this.generateCards(this.props.reportOnMarkets, 'reportOnMarkets'))
-      .concat(this.generateCards(this.props.finalizedMarkets, 'finalizedMarkets')));
+      this.generateCards(props.resolvedMarketsOpenOrders, 'resolvedMarketsOpenOrders')
```
bthaile: 'resolvedMarketsOpenOrders' like these can be put in constants as NOTIFICATION_TYPES: {....}

I'd like to remove magic strings.

src/modules/notifications/selectors/notification-state.js
```diff
@@ -49,6 +49,51 @@ export const selectFinalizeMarkets = createSelector(
...
+  }
+);
+
+// Get all markets in dispute, for market creators and user with positions in disputed markets
+export const selectMarketsInDispute = createSelector(
+  selectMarkets,
+  selectLoginAccountAddress,
+  (markets, address) => {
+    if (markets.length > 0) {
+      return markets
```
bthaile: are all markets in dispute been loaded?

I think there is a call to get all markets that user has staked on, that would give us the marketIds then we can load them.

src/modules/account/components/notifications/notifications.tsx
andrewdamelio: 👍 on it

src/modules/notifications/selectors/notification-state.js
bthaile: ok, understood on needing to loadDisputing

andrewdamelio: ⚡️

PR #2724, opened 2019-03-07
Clean up market selector

bthaile:
slim up market selector in effect to reduce complexity and reduce frequency it's triggered. 

This is first pass to help diagnose why market selector gets triggered frequently. Seems each block the market selector is triggered per market, not consistently though.

also, changed default value for Has Open Orders to false. needed to be done


src/modules/markets/selectors/market.js
```diff
@@ -78,505 +69,389 @@ import { selectReportableOutcomes } from "modules/reports/selectors/reportable-o
...
+    });
+
+    if (
+      marketAccountTrades ||
+      (marketAccountPositions &&
+        marketAccountPositions.tradingPositionsPerMarket)
+    ) {
+      market.myPositionsSummary = {};
+      const marketPositions = marketAccountPositions.tradingPositionsPerMarket;
+      // leave complete sets here, until we finish new notifications
```
andrewdamelio: 👍

PR #2727, opened 2019-03-08
add notifications to claim reporting fees and cancel/sign unsigned orders

andrewdamelio:
### Add notifications 

- added a notification to claim reporting fees
- added a notification to cancel/sign unsigned orders
- refactored / moved component logic to selectors

### DIFF
--- a/src/modules/notifications/selectors/notification-state.js
+++ b/src/modules/notifications/selectors/notification-state.js
@@ -1,8 +1,25 @@
 import { createSelector } from "reselect";
 import { selectMarkets } from "src/modules/markets/selectors/markets-all";
-import { selectLoginAccountAddress } from "src/select-state";
-import { MARKET_CLOSED } from "src/modules/common-elements/constants";
+import {
+  selectLoginAccountAddress,
+  selectReportingWindowStats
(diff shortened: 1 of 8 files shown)


src/modules/notifications/selectors/notification-state.js
```diff
@@ -100,6 +117,113 @@ export const selectMarketsInDispute = createSelector(
...
+  }
+);
+
+// Get all unsigned orders from localStorage
+export const selectUnsignedOrders = createSelector(
+  selectMarkets,
+  selectLoginAccountAddress,
+  (markets, userAddress) => {
+    if (localStorageRef && userAddress) {
+      const getUserState = JSON.parse(localStorageRef.getItem(userAddress));
```
bthaile: there is a reducer that manages this data.

`loadPendingLiquidityOrders` in src/modules/auth/actions/load-account-data-from-local-storage.js

Also, the modal will clear liquidity orders depending if user cancels or sends. Also the direction is to consolidate interaction with localStorage.

andrewdamelio: fixed! using the selector for this now.

PR #2736, opened 2019-03-12
handle last open order being canceled when on portfolio page.

bthaile:
fixes https://github.com/AugurProject/augur/issues/1288

simple solution, but we need to think about it more. There might be more situations we need to handel the last item of a market being removed from the quads.


bthaile:
There is still an issue if user has multiple open orders in many markets and cancels the last one in a single market. Still working

PR #2751, opened 2019-03-14
use platform time frame data

phoebemirman:
fixes https://github.com/AugurProject/augur/issues/1193


src/modules/account/components/augur-status/activity.tsx
```diff
@@ -20,8 +22,15 @@ export default class Activity extends React.Component<
     selected: constants.TIMEFRAME_OPTIONS[0].id
   };
 
+  componentDidMount() {
+    this.props.updatePlatformTimeframeData({
+      endTime: constants.TIMEFRAME_OPTIONS[0].id
```
bthaile: need to calculate endTime to send to augur-node.  I'll move how PnL chart does it to a helper

PR #2758, opened 2019-03-18
Countdown timer proceeds

andrewdamelio:
### Countdown timer proceeds + some other minor fixes
- Added two types of Claim Proceeds notification. 
      -- One for ready to claim rewards
      -- One for rewards that have to wait for a timer
Screen Shot 2019-03-15 at 11 32 54 AM
Screen Shot 2019-03-15 at 11 37 45 AM

- Refactored Notifications in preparation for New/Used state plus general cleanup.

- Added empty state for notifications
Screen Shot 2019-03-17 at 11 54 44 AM

- Claim Proceeds modal will **now** close and not throw an error on Market Header
Screen Shot 2019-03-14 at 10 26 34 AM

- Wired up modal
Screen Shot 2019-03-15 at 11 33 01 AM


src/modules/notifications/selectors/notification-state.js
```diff
@@ -145,6 +155,43 @@ export const selectProceedsToClaim = createSelector(
...
+// Get all markets where the user has outstanding returns but needs to wait CLAIM_PROCEEDS_WAIT_TIME
+export const selectProceedsToClaimOnHold = createSelector(
+  selectAllProceedsToClaim,
+  selectCurrentTimestampInSeconds,
+  (markets, currentTimestamp) => {
+    if (markets.length > 0 && currentTimestamp) {
+      return markets
+        .filter(
+          market =>
+            !canClaimProceeds(
```
bthaile: Looks like this won't return markets where the waiting period hasn't passed.

don't you want all markets where user has winning shares `outstandingReturns`

andrewdamelio: I created three selectors to manage this.

First ```selectAllProceedsToClaim ``` gets all finalized markets with outstanding shares.

Second ```selectProceedsToClaim``` gets all markets that are claimable

and finally ```selectProceedsToClaimOnHold``` gets all markest that need to wait until claimable 

Really ```selectProceedsToClaim``` and ```selectProceedsToClaimOnHold``` are just the same but inverted.

bthaile: // Get all markets where the user has outstanding returns but needs to wait 

this comment, It seems you don't want to filter out markets that user has winning shares, but will need to wait.

andrewdamelio: Correct, for this selector, I don't want to filter out markets that an user has winning shares for but has it wait. 

This selector does the oppsoite, it filters out the markets that an user has wining share for and are ready to be claim right now.

I do this by grabbing all the markets with winning shares and filtering on all the markets that **can't** canClaimProceeds yet.

```
        .filter(
          market =>
            !canClaimProceeds(...
```

PR #2764, opened 2019-03-18
add leading zeros

phoebemirman:
https://github.com/AugurProject/augur/issues/1371


src/utils/format-date.js
```diff
@@ -63,7 +63,7 @@ export function formatDate(d) {
     } ${date.getUTCDate()}, ${date.getUTCFullYear()} ${utcTimeTwelve.join(
       ":"
     )} ${utcAMPM}`, // UTC time
-    formattedShortDate: `${date.getUTCDate()}${
+    formattedShortDate: `${addLeadingZero(date.getUTCDate())}${
```
bthaile: we already add leading zeros for other values like hours, minutes, seconds. use the same format. Just for consistency.

`    ("0" + date.getUTCHours()).slice(-2),`

PR #2785, opened 2019-03-20
Remove price next to scalar denomination

andrewdamelio:
image


src/modules/market/components/market-outcomes-list--outcome/market-outcomes-list--outcome.jsx
```diff
@@ -49,9 +48,6 @@ const Outcome = ({
             {outcomeName || (scalarDenomination && scalarDenomination)}{" "}
           </span>
         </div>
```
bthaile: We only want to remove the percentage for scalars, not all market types.

andrewdamelio: oh, good catch! my mistake

andrewdamelio: updated

Issue #2772, opened 2019-03-19
hover tooltip numeric showing too many decimals


bthaile:
tooltip needs to only show 3 decimals with ellipse. This is causing table row values to be too long when space is tight. 

`0.00001113`
should show `0.000...` as an example.

Chwy5:
reviewed - looks correct - closing

PR #2801, opened 2019-03-21
updated labels to always be upper case text for common elements

JohnDanz:
https://github.com/AugurProject/augur/issues/1398
https://github.com/augurproject/augur/issues/1399
https://github.com/augurproject/augur-ui/issues/2772

https://github.com/AugurProject/augur/issues/1419

I think that's all the issues covered in this


bthaile:
There are other places where uppercase needs to take place, tix will be created as needed.

PR #2815, opened 2019-03-22
style open order rows for trading page

phoebemirman:
https://github.com/AugurProject/augur/issues/1440

styled open order rows to match the mock for trading page and updated portfolio with the new styles


bthaile:
Warning needs to be cleaned up

image

Issue #2773, opened 2019-03-19
portfolio: Filled order quad text sizing


bthaile:
existing tix for numeric values for portfolio.

Also whole table row in filled orders needs to be 11px.

Market title needs to be 12px.

Chwy5:
@matt-bullock please review and close if this looks correct

matt-bullock:
outcome should be medium in weight (500)

bthaile:
@phoebemirman comment on this issue.

Chwy5:
@matt-bullock double check this is correct and close if it is

matt-bullock:
closing

PR #2818, opened 2019-03-25
Trading orders mobile

phoebemirman:
https://github.com/AugurProject/augur/issues/1466
https://github.com/AugurProject/augur/issues/1465
https://github.com/AugurProject/augur-ui/issues/2773


bthaile:
good call on the caps

PR #2820, opened 2019-03-25
Trading page mobile

phoebemirman:
https://github.com/AugurProject/augur/issues/1468

https://github.com/AugurProject/augur/issues/1467


src/modules/market/components/market-view/market-view.styles.less
```diff
@@ -218,4 +230,8 @@
   .MarketView {
     height: 100%;
   }
+
+  .MarketView__container {
```
bthaile: Could just remove this class

PR #2822, opened 2019-03-26
Remove timeouts

bthaile:
removed code no longer needed.

We are moving away form font awesome, there were many setTimeout violations when rocket-loader was trying to pull in font awesome.

Also removed a setTimeout having to do with setting market not loading, missed those on previous refactor.

https://github.com/AugurProject/augur/issues/1479


src/modules/common/components/input-dropdown/input-dropdown.jsx
```diff
@@ -125,12 +125,11 @@ class InputDropdown extends Component {
...
         </select>
-        <i
-          className={classNames(Styles.InputDropdown__icon, "fa", {
-            "fa-angle-down": !this.state.showList,
-            "fa-angle-up": this.state.showList
-          })}
-        />
+        <span className={Styles.InputDropdown__icon}>
+          {!this.state.showList
+            ? ChevronDown({ stroke: "#FFFFFF" })
```
phoebemirman: can you use the ChevronFlip component for this ? Then you won't have to reuse code.

bthaile: updated two places to use ChevronFlip.

src/modules/forking/components/forking-alert/forking-alert.styles.less
```diff
@@ -49,6 +49,15 @@
 
 .ForkingAlert__arrow {
   margin-left: 1em;
+
+  svg {
+    height: 1em;
+    margin-right: 0.5em;
+
+    polyline {
+      stroke: @color-white;
```
phoebemirman: do you still need this here and in the other spot?

src/modules/common/components/input-dropdown/input-dropdown.styles.less
```diff
@@ -73,6 +73,14 @@
 
 .InputDropdown__icon {
   align-self: center;
```
phoebemirman: this should be flex-end so the arrow is aligned

bthaile: made the correction.

bthaile:
Fixed search functionality.

PR #2824, opened 2019-03-26
Create selector market

bthaile:
moved dispatched method out of market selector so dispatch isn't needed to be passed in.

set up simple selectors to return market specific data from state collections.

This is a jumping off point for more complex ways to do cacheing of selectors. I restructured the market selector, not much of a performance boost, most improvement is the selector actually stops triggering if the UI just sits there, until a new block comes in. so I don't think we have big leaks

continued performance task and when user connects
https://github.com/AugurProject/augur/issues/1284

### DIFF
--- a/src/modules/markets/selectors/market.js
+++ b/src/modules/markets/selectors/market.js
@@ -164,8 +164,6 @@ const assembleMarket = (
   pendingOrders
 ) => {
   const marketId = marketData.id;
-
-  console.log("trigger market selector", marketId);
   const market = {
     ...marketData,
     description: marketData.description || "",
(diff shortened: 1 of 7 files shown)


src/modules/markets/selectors/market.js
```diff
@@ -67,45 +64,72 @@ import { selectReportableOutcomes } from "modules/reports/selectors/reportable-o
...
+const selectOrderCancellationStateMarket = (state, marketId) =>
+  selectOrderCancellationState(state)[marketId] || {};
+
+const selectAccountShareBalanceMarket = (state, marketId) =>
+  selectAccountShareBalance(state)[marketId];
+
+const selectPendingOrdersStateMarket = (state, marketId) =>
+  selectPendingOrdersState(state)[marketId];
+
+const getMarketSelector = createSelector(
```
andrewdamelio: 👍

PR #2830, opened 2019-03-27
Timezone banner tooltip

bthaile:
https://github.com/AugurProject/augur/issues/1486

When user logs in the market banner is dismissible.


src/modules/market/components/market-header-message/market-header-message.jsx
```diff
@@ -0,0 +1,88 @@
...
+            >
+              <div
+                style={{ color: "#372e4b" }}
+                className={Styles.MarketHeaderMessage__tooltip}
+              >
+                {
+                  <div>
+                    <p>
+                      Markets on Augur are created by the community, this means
+                      that errors are can be made in the creation of a market,
```
phoebemirman: "are" and "can" should not both be here. get rid of "are"

src/modules/market/components/market-header-message/market-header-message.jsx
```diff
@@ -0,0 +1,88 @@
...
+                      which can result in the market being resolved as invalid.
+                      If a market resolves as invalid, each share is refunded to
+                      traders in equally amounts.
```
phoebemirman: should be "equal"

src/modules/market/components/market-header-message/market-header-message.jsx
bthaile: `  that errors can be made in the creation of a market,` this does read better. good one

src/modules/market/components/market-header-message/market-header-message.jsx
bthaile: fixed

PR #2865, opened 2019-04-01
Timezone change market creation

bthaile:
https://github.com/AugurProject/augur/issues/1516

first installment, I'd like to get these changes up on dev.augur.net for design to see. There are questions about mobile and what the dispute form should look/behave like when user fills dispute bond.

### DIFF
--- a/src/modules/reporting/components/common/invaild-message.styles.less
+++ b/src/modules/reporting/components/common/invaild-message.styles.less
@@ -4,8 +4,8 @@
   color: @color-darkpurple;
   display: flex;
   flex-direction: column;
-  font-size: @font-size-medium;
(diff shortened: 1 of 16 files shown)


src/modules/reporting/components/common/invaild-message.styles.less
```diff
@@ -0,0 +1,30 @@
...
+
+    > span {
+      &.bolden {
+        font-weight: bold;
+      }
+    }
+  }
+}
+
+.InvaildMessage li {
```
phoebemirman: it is spelled as "invalid"

bthaile: updated to fix the typo

PR #2902, opened 2019-04-11
switch my reporting to reporting tab

phoebemirman:
https://github.com/AugurProject/augur/issues/1636


src/modules/routes/constants/views.js
```diff
@@ -18,11 +18,8 @@ export const DEFAULT_VIEW = MARKETS;
 
 // SUB VIEWS
 //  Portfolio
-export const MY_POSITIONS = "my-positions";
-export const MY_MARKETS = "my-markets";
-export const FAVORITES = "favorites";
+export const MY_POSITIONS = "portfolio";
 export const PORTFOLIO_TRANSACTIONS = "transactions";
```
bthaile: can `PORTFOLIO_TRANSACTIONS` be removed as well?

phoebemirman: yea true

PR #2916, opened 2019-04-15
add-rep-balance-component

JohnDanz:
https://github.com/AugurProject/augur/issues/1643


bthaile:
There are a few reporting changes we need to migrate from master. 

master's trading page changes we don't want, 
market creation changes we don't want. 

I'll re-create the reporting changes on new-trading-page-layout.

PR #2941, opened 2019-04-17
update totals when claiming individually

phoebemirman:
https://github.com/AugurProject/augur/issues/1623

### DIFF
--- a/src/modules/modal/containers/modal-claim-fees.js
+++ b/src/modules/modal/containers/modal-claim-fees.js
@@ -8,7 +8,8 @@ import {
   formatGasCostToEther,
   formatAttoRep,
   formatAttoEth,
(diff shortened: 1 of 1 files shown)


src/modules/modal/containers/modal-claim-fees.js
```diff
@@ -169,11 +203,11 @@ const mergeProps = (sP: any, dP: any, oP: any) => {
       ? [
           {
             label: "Total REP",
-            value: `${sP.reportingFees.unclaimedRep.full} REP`
+            value: `${formatRep(unclaimedRep.toNumber()).full} REP`
```
bthaile: FYI, format-number can take number, string or BigNumber no need to call `.toNumber()`.

PR #2943, opened 2019-04-18
Reporting timezone changes

bthaile:
first installment of re-creating changes on dev branch from master for 

https://github.com/AugurProject/augur/issues/1686

changes:
added yellow information bubble thing on reporting and disputing. Changed up styling for mobile.
Fixed bug where reporting form was getting wrong initial reporting stake for designated reporter.
Now show amount REP needed for Designated reporter even if the market is in open reporting.

image

image

image

image

### DIFF
--- a/src/assets/styles/_colors_redesign.less
+++ b/src/assets/styles/_colors_redesign.less
@@ -13,7 +13,7 @@
 @color-secondary-background: #120D1B;
 @color-tertiary-background: #1B1528;
 @color-theme-primary: #211A32;
-@color-yellowish-alert: #FBE9A6;
+@color-info-message: #FBE9A6;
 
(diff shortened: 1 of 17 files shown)


src/assets/styles/_colors_redesign.less
```diff
@@ -13,16 +13,18 @@
 @color-secondary-background: #120D1B;
 @color-tertiary-background: #1B1528;
 @color-theme-primary: #211A32;
+@color-yellowish-alert: #FBE9A6;
```
phoebemirman: don't think you are supposed to have color names in here?

bthaile: Hows that. Changed to info-message

PR #2944, opened 2019-04-18
design updates - market header

andrewdamelio:
https://github.com/AugurProject/augur/issues/1535

> Looks good. There are few things that still need addressing. Nothing major but would be good to get these in.
> 
> * [x]  update this arrow to the arrow we're using across the designs
>   
> * [x]  Nudge up roughly 2px the Estimated fee figure - so it lines up with 24hr volume figure
>   
> * [x]  Can we add some line height to the additional details copy. I have 140% in the figma file. This should be standard for 12px Roboto paragraph text. Lets it breathe and is easier to read
>   
> * [x]   update 'More' button - see design
> * [x]  hidden text should end with ellipses and shouldn't cut through a line of text. This is what I'm seeing on the build. Is it possible for it to work like the screenshot above?
>   
> * [x]  animation to show less/more could be smoothed out. When I click 'less' it pushes the content below down before pulling it back up. Should be a smooth eased transition with no jerkiness.
> * [x]  default state for menu button should be #76737E. White on hover/active
>   
> * [x]  change timeframe line background colour to #76737E
>

### DIFF
--- a/src/modules/market/components/market-header/market-header.jsx
+++ b/src/modules/market/components/market-header/market-header.jsx
@@ -30,7 +30,8 @@ import { MarketTimeline } from "modules/common-elements/progress";
 
 import ToggleHeightStyles from "utils/toggle-height/toggle-height.styles";
 
-const OVERFLOW_DETAILS_LENGTH = 89; // in px, matches additional details label max-height
+const MAX_DETAILS_LENGTH = 500; // Show MORE for details when greater than 500 char length
+
 export default class MarketHeader extends Component {
   static propTypes = {
     description: PropTypes.string.isRequired,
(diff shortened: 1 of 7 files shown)


src/modules/market/components/market-header/market-header.jsx
```diff
@@ -373,7 +369,12 @@ export default class MarketHeader extends Component {
               this.toggleMarketHeader(currentAugurTimestamp, market)
             }
           >
-            <ChevronFlip quick pointDown={headerCollapsed} stroke="white" />
+            <ChevronFlip
+              stroke="#999999"
```
bthaile: could use `@color-neutral` here. I'll change before merging to sneakpreview

PR #2954, opened 2019-04-20
performance and optimizations improvements

bthaile:
https://github.com/AugurProject/augur/issues/1284

*  added re-reselect, which caches the market selector by marketId
* refactored how market orders is populated in state, to reduce trips to augur-node
* removed unneeded reducer assoc. with orders
* removed call to loadDisputes in markets-view (main listing of markets) it was triggering market selector. loadDisputes is called as part of load market if not already loaded when the specific market is nav to.


src/modules/app/actions/sync-blockchain.js
```diff
@@ -40,18 +38,15 @@ export const syncBlockchain = cb => (dispatch, getState) => {
     cb && cb();
   });
 
-  debounce(
```
justinbarry: Should we be debouncing this?

bthaile: Dont need it. Sync blockchain is called every block. I recently added it b/c my debug logging suggested it was called twice on occasion. 

What was really going on was augur-node takes longer every second get sync data request. So the responses look like a double tap, but one of them was just lagging.

andrewdamelio:
+1 on re-reselect Tom! nice

PR #2977, opened 2019-04-24
update mobile your overview

andrewdamelio:
https://github.com/AugurProject/augur/issues/1741

## OLD
Screen Shot 2019-04-24 at 11 56 10 AM

## NEW
Screen Shot 2019-04-24 at 11 37 48 AM

### DIFF
--- a/src/modules/account/components/overview-stats/overview-stats.styles.less
+++ b/src/modules/account/components/overview-stats/overview-stats.styles.less
@@ -9,6 +9,19 @@
     margin-bottom: 0.375rem;
   }
 
+  @media @breakpoint-mobile {
+    > div {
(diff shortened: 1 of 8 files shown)


src/modules/account/components/overview-stats/overview-stats.styles.less
```diff
@@ -9,6 +9,19 @@
     margin-bottom: 0.375rem;
   }
 
+  @media @breakpoint-mobile {
+    > div {
+      > span:first-of-type {
+        font-size: @size-10;
```
JohnDanz: use the saved font-styles here instead of directly changing font-size, same for below (the .text/.mono stuff)

PR #2978, opened 2019-04-24
reflow create market mobile view to column

JohnDanz:
made a few cleanups to create-market but it really needs to be refactored. reflowed the view on mobile to a column style for now so we can get 1.13 out the door

https://github.com/AugurProject/augur/issues/1845


bthaile:
image

bthaile:
can you see if you can fix this white button thing. Not sure what is going on there. 

there is a svg error, but I don't think that has anything to do with it. 
Screen Shot 2019-04-24 at 3 33 28 PM


image

PR #2984, opened 2019-04-25
More Responsive fixes

andrewdamelio:
https://github.com/AugurProject/augur/issues/1641
**At 1280px I'm also seeing a problem with uneven column widths:**
- Until we remove the side nav, to address the uneven box widths, we've decided to adjust the font size of the switch label component between 1281px and 1420px.

Screen Shot 2019-04-25 at 11 09 44 AM

### DIFF
--- a/src/modules/common-elements/switch-labels-group.styles.less
+++ b/src/modules/common-elements/switch-labels-group.styles.less
@@ -49,6 +49,12 @@
     }
   }
 
+  // TODO: Circle back and remove this when SIDE NAV is removed
+  // Temporary quick fix for sizing issues between 1281 - 1420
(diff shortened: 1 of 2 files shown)


src/modules/common-elements/switch-labels-group.styles.less
```diff
@@ -49,6 +49,12 @@
     }
   }
 
+  // TODO: Circle back and remove this when SIDE NAV is removed
+  // Temporary quick fix for sizing issues between 1281 - 1420
+  @media (min-width: 1281px) and (max-width: 1420px) {
```
JohnDanz: make a breakpoint for this, call it temp or something so we know to fix it later, remove TODO

PR #2991, opened 2019-04-25
Filled orders fixes

phoebemirman:
https://github.com/AugurProject/augur/issues/1878
https://github.com/AugurProject/augur/issues/1879

re-update search less often when switching views


bthaile:
@JohnDanz Here are some changes to common components. Do you want to take a look?

PR #3005, opened 2019-04-29
Modal Scroll

JohnDanz:
scrolling to footer fixed, cleaned up some small issues i was seeing on smaller sizes for the modals

https://github.com/AugurProject/augur/issues/1601


src/modules/modal/modal.styles.less
```diff
@@ -47,7 +46,7 @@
   > div:last-of-type {
     display: flex;
     flex-flow: row-reverse nowrap;
```
andrewdamelio: Can we remove instead of comment?

PR #3008, opened 2019-04-29
fix filled order sorting

phoebemirman:
https://github.com/augurproject/augur/issues/1938
before fix:
Screen Shot 2019-04-29 at 11 41 26 AM

after fix:
Screen Shot 2019-04-29 at 11 42 32 AM

### DIFF
--- a/src/assets/styles/_responsive.less
+++ b/src/assets/styles/_responsive.less
@@ -76,4 +76,4 @@
 @breakpoint-mobile-extra-small-height: ~"all and (max-height: @{step-4})";
 
 // TODO remove when side nav is removed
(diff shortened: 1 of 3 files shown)


src/assets/styles/_responsive.less
```diff
@@ -76,4 +76,4 @@
 @breakpoint-mobile-extra-small-height: ~"all and (max-height: @{step-4})";
 
 // TODO remove when side nav is removed
-@breakpoint-temp: ~"(min-width: 1281px) and (max-width: 1420px)";
+@breakpoint-temp: ~"all and (min-width: 1281px) and (max-width: 1420px)";
```
bthaile: I don't think this is needed, I don't want to mod breakpoints at this time

PR #3021, opened 2019-04-30
order book was flickering when log came in, issue was old way of tryi…

bthaile:
…ng to update an idividual order and how we get all open orders to update the order book

https://github.com/AugurProject/augur/issues/1953

removed the order book actions that haven't been used. Also remove updateOrder since we are getting all open orders if user is watching the specific market. otherwise we are only interested in open orders if user's open order gets created/canceled/filled.


src/modules/orders/reducers/order-books.js
```diff
@@ -10,6 +10,7 @@ export default function(orderBooks = DEFAULT_STATE, { type, data }) {
   switch (type) {
     case UPDATE_ORDER_BOOK: {
       const { marketId, orderBook } = data;
+      console.log("order book", marketId, orderBook);
```
phoebemirman: should this still be here?

bthaile: I'll pull that out. oops

PR #3039, opened 2019-05-06
market creation cutoff updates

phoebemirman:
add inner banner to app to be used for reporting/disputing, trading pg banners, add it for market creation, add message on market creation, disable dates past cutoff on market creation

https://github.com/AugurProject/augur/issues/1986

screenshots:

Screen Shot 2019-05-06 at 12 19 05 PM
Screen Shot 2019-05-06 at 12 18 50 PM
Screen Shot 2019-05-06 at 12 18 43 PM

### DIFF
--- a/src/modules/app/components/app/app.jsx
+++ b/src/modules/app/components/app/app.jsx
@@ -26,6 +26,7 @@ import SideNav from "modules/app/components/side-nav/side-nav";
 import Logo from "modules/app/components/logo/logo";
 import Routes from "modules/routes/components/routes/routes";
 import NotificationsContainer from "modules/notifications/containers/notifications-view";
+import InnerBanner from "modules/app/components/inner-banner/inner-banner";
 
 import MobileNavHamburgerIcon from "modules/common/components/mobile-nav-hamburger-icon";
 import MobileNavCloseIcon from "modules/common/components/mobile-nav-close-icon";
@@ -44,6 +45,9 @@ import parseQuery from "modules/routes/helpers/parse-query";
 import getValue from "utils/get-value";
 
 import {
+  // MARKET,
+  // REPORT,
+  // DISPUTE,
   MARKETS,
   ACCOUNT_DEPOSIT,
   ACCOUNT_WITHDRAW,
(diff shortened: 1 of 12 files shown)


src/modules/create-market/components/create-market-form-define/create-market-form-define.jsx
```diff
@@ -174,7 +175,25 @@ export default class CreateMarketDefine extends Component {
...
+              Styles.CreateMarketDefine_message,
+              Styles.CutoffMessage
+            )}
+          >
+            <div>
+              No new markets can be created that end after{" "}
+              <span>{CUTOFF_READABLE}</span>. Any markets that end after this
+              date cannot be guaranteed to resolve correctly.{" "}
+              <a
+                href="http://docs.augur.net"
```
andrewdamelio: Is the **Read More** suppose to go http://docs.augur.net or is that just temporary until we have confirmed documentation to provide?

Is it worth adding this as a constant for easy updating later?

src/modules/app/components/app/app.jsx
```diff
@@ -44,6 +45,9 @@ import parseQuery from "modules/routes/helpers/parse-query";
 import getValue from "utils/get-value";
 
 import {
+  // MARKET,
+  // REPORT,
+  // DISPUTE,
```
andrewdamelio: Any reason we can't remove these instead of commenting them out?

src/modules/app/components/app/app.jsx
```diff
@@ -538,13 +542,19 @@ export default class AppView extends Component {
       tagsMargin = 110 * subMenu.scalar;
     }
 
+    const showBanner = currentPath === CREATE_MARKET;
+    // currentPath === MARKET ||
```
andrewdamelio: Same here

src/modules/app/components/app/app.jsx
phoebemirman: i added those comments for the other banners that are going to use that. but i can get rid of them in the meantime.

PR #3091, opened 2019-06-14
Fix x trading banner

bthaile:
clicking on [x] on market trading banner doesn't do anything. Here is the fix


bthaile:
ok, I see. Don't show the message if market's endTime is beyond v2 cutoff.

PR #3099, opened 2019-07-18
merge categories around, only clear when no category is passed in

bthaile:
keep categories state and only clear if category (remove categories collection in redux state) isn't passed into query filter. The effect is the side categories are not cleared if the user clicks on a category. 

image

### DIFF
--- a/src/modules/categories/actions/load-categories.js
+++ b/src/modules/categories/actions/load-categories.js
@@ -5,10 +5,11 @@ import {
 } from "modules/categories/actions/update-categories";
 import logError from "utils/log-error";
 
-const loadCategories = (params = {}, callback = logError) => (
-  dispatch,
-  getState
-) => {
+const loadCategories = (
+  params = {},
+  refreshCategories,
+  callback = logError
+) => (dispatch, getState) => {
   const { universe } = getState();
   if (!universe.id) return callback(null);
   augur.markets.getCategories(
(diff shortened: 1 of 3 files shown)


src/modules/markets/actions/load-markets.js
```diff
@@ -197,7 +197,7 @@ export const loadMarketsByFilter = (filterOptions, cb = () => {}) => (
     }, 2000);
 
     // load categories here
-    dispatch(loadCategories(params));
+    dispatch(loadCategories(params, !params.category));
```
bthaile: This looks if a category is included in the filters. if it is tell redux store not to clear, just merge.

src/modules/categories/reducers/categories-data.js
```diff
@@ -10,7 +10,17 @@ const DEFAULT_STATE = [];
...
+        new Set([
+          ...categories.map(c => c.categoryName),
+          ...data.categories.map(c => c.categoryName)
+        ])
+      ).map(
+        k =>
+          data.categories.find(d => d.categoryName === k)
+            ? data.categories.find(d => d.categoryName === k)
+            : categories.find(d => d.categoryName === k)
+      );
```
bthaile: logic to merge in category objects into existing collection

src/modules/categories/actions/load-categories.js
```diff
@@ -5,18 +5,19 @@ import {
...
+) => (dispatch, getState) => {
   const { universe } = getState();
   if (!universe.id) return callback(null);
   augur.markets.getCategories(
     { ...params, universe: universe.id },
     (err, categories) => {
       if (err) return callback(err);
       if (categories == null) return callback(null);
-      dispatch(clearCategories());
+      if (refreshCategories) dispatch(clearCategories());
```
bthaile: refresh meaning clear redux state collection.

PR #3116, opened 2019-09-14
Remove post v2 cutoff filter

bthaile:
Remove past v2 filter checkbox

image

### DIFF
--- a/src/modules/filter-sort/components/filter-dropdowns/filter-dropdowns.jsx
+++ b/src/modules/filter-sort/components/filter-dropdowns/filter-dropdowns.jsx
@@ -97,7 +97,6 @@ export default class FilterSearch extends Component {
     updateHideInsecureMarkets: PropTypes.func.isRequired,
     history: PropTypes.object.isRequired,
     location: PropTypes.object.isRequired,
-    hasPositionsInCutoffMarkets: PropTypes.bool.isRequired,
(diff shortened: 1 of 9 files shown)


src/modules/filter-sort/components/filter-dropdowns/filter-dropdowns.jsx
```diff
@@ -111,22 +108,10 @@ export default class FilterSearch extends Component {
...
     this.changeMaxSpread = this.changeMaxSpread.bind(this);
     this.goToPageOne = this.goToPageOne.bind(this);
-    this.changeHidePastCutoff = this.changeHidePastCutoff.bind(this);
     this.changeExperimentalInvalid = this.changeExperimentalInvalid.bind(this);
     this.changeHideInsureMarkets = this.changeHideInsureMarkets.bind(this);
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (
-      nextProps.hasPositionsInCutoffMarkets !==
```
pgebheim: Oh I bet the new filter won't show markets you have positions in if they're past they don't have enough IR stake.

This but reminded me of that logic.

bthaile: Yea, we just want to show the pop up modal. I would think.  I do not know what the model says but might as will pop up

PR #3129, opened 2019-10-29
Removes Website wording in resolution source during market creation.

MicahZoltu:


MicahZoltu:
Pretty sure this will break any tests, but wanted to get the proposal out there and started.  Someone who actually knows what they are doing, feel free to take this over.  :)

MicahZoltu:
Why didn't this fail?  https://github.com/AugurProject/augur-ui/blob/master/integration/create-market.test.ts#L158  Are tests not run as part of CI?

bthaile:
I think there is a market creation test that needs to be updated. And look, there might be another instance of that phrase in the market creation code.

nuevoalex:
I don't think we have any tests that just check if all copy is what we set it to. In general any test I'd want a script to update for me isn't very useful

pgebheim:
Those tests (the integration ones) don't run in CI because they're so flakey.

